PR #26, opened 2017-07-03
Floor ceiling

callahanhirrel:


byorgey:
See, I knew you could knock it out in an afternoon. =)

Overall this looks great.  Let's also add a new test case for these operators.  I also added a few comments inline.

src/Disco/Typecheck.hs
```diff
@@ -463,6 +469,18 @@ infer (TUn Lg t)  = do
   at <- check t TyN
   return $ ATUn TyN Lg at
 
+infer (TUn Floor t) = do
+  at <- infer t
+  checkNumTy at
+  -- ty <- convNumTy (getType at) why does this cause an error?
```
byorgey: Remember that `ty <- convNumTy (getType at)` is syntax sugar for  `convNumTy (getType at) >>= \ty -> ...`.  Recall also that `>>=` has a type like `m a -> (a -> m b) -> m b`.  So its first argument (in this case `convNumTy (getType at)`) has to be a monadic computation, in this case it would have to have a type like `TCM Type`.  But it actually has the plain type `Type`.  If you want to just assign a name to a plain value inside a `do`-block, you can use `let name = value`, as in
```
infer (TUn Floor t) = do
  at <- infer t
  checkNumTy at
  let ty = convNumTy (getType at)
  return $ ATUn ty Floor at
```
It's up to you whether you want to use a `let` in this way or if you just want to delete the comment.

src/Disco/Typecheck.hs
```diff
@@ -463,6 +469,18 @@ infer (TUn Lg t)  = do
...
+  return $ ATUn (convNumTy (getType at)) Floor at
+
+infer (TUn Ceil t) = do
+  at <- infer t
+  checkNumTy at
+  -- ty <- convNumTy (getType at)
```
byorgey: Same thing here of course.

src/Disco/Interpret/Core.hs
```diff
@@ -457,6 +459,12 @@ integerSqrt' n =
 (^!) :: Num a => a -> Int -> a
 (^!) x n = x^n
 
+floorOp :: Rational -> Rational
+floorOp n = (floor n) % 1
+
+ceilOp :: Rational -> Rational
+ceilOp n  = (ceiling n) % 1
```
byorgey: Perfect.

byorgey:
Oh, and also, did you want to take a shot at adding ⌊x⌋ and ⌈x⌉ syntax?

Issue #10, opened 2017-06-19
Integer division operator


byorgey:
Let's add a new binary operator `//` (a la Python) which does integer division.  Technically this is unnecessary once we have `floor` --- and indeed we can treat `x // y` as syntax sugar for `floor (x / y)`---but I think it will still be nice to have it in the language explicitly.

The typing rule looks a bit complicated but it follows from the typing rules for `/` and `floor`:
```
Γ ⊢ t1 ▷ ν1      Γ ⊢ t2 ▷ ν2    ν3 = ν1 ⊔ ν2
--------------------------------------------------------
Γ ⊢ t1 // t2 ▷ int(ν3)
```
This says that if we can infer that t1 has type ν1 and t2 has type ν2 (both of which must be numeric types), then `t1 // t2` has type int(ν3), where `int` is the function on types defined in #9 , and ν3 is the least upper bound (in the little diamond lattice of numeric types) of ν1 and ν2. (If we directly compose the typing rules for `/` and `floor` we might expect to have ν3 = ν1 ⊔ ν2 ⊔ Q+ --- but int(ν1 ⊔ ν2 ⊔ Q+) = int(ν1) ⊔ int(ν2) ⊔ int(Q+) = int(ν1) ⊔ int(ν2) ⊔ N = int(ν1) ⊔ int(ν2) = int(ν1 ⊔ ν2).)  You can start by copying the code for type inference on the division operator --- the only difference for `//` will be the `int` at the end.

After implementing parsing and typechecking, you can add a rule to the desugarer which just desugars `x // y` to `floor (x / y)` and then you shouldn't need to touch the interpreter.

byorgey:
Closed by #37.

PR #37, opened 2017-07-06
Integer division

callahanhirrel:


callahanhirrel:
Closes #10

byorgey:
Incidentally, adding a comment like "Closes #10" to the *pull request* doesn't work.  It has to be in a comment on one of the commits.

heades:
A pull request will trigger a commit, but if you add:

close #10

to the commit message, then it will close the issue.

PR #40, opened 2017-07-07
Put all operator info (fixity, precedence, concrete syntax...) in one place

### LINKED ISSUE #35: fromJust Nothing
```
Disco> :type (1 :: [4,5])
disco: Maybe.fromJust: Nothing
```
Just discovered this --- not sure what is causing it.  Will look into it more later.

byorgey:
Fixes #35.


byorgey:
The tests will probably fail because of the divides operator, which still needs to be changed in one of the tests.

PR #45, opened 2017-07-10
List comprehensions

byorgey:
It was actually quite difficult to figure out how to implement list comprehensions such that they would be sufficiently lazy.  In the end, the trick is to add a new `VDelay` constructor to the `Value` type, which contains a (delayed) `IM Value` computation which can be executed later, and then to use this to implement various "list" functions (like `append`, `concat`, and `map`) on `Value`s representing disco lists, but in such a way that the functions will not evaluate at all until their results are actually demanded by `whnf`.


callahanhirrel:
Looks good!

PR #48, opened 2017-07-11
Switch from unbound to unbound-generics

### LINKED ISSUE #44: Shadowing in list comprehensions
```
Disco> :type [ x | x <- [1,2,3], x <- [4,5,6]]
[x | x in [1, 2, 3], x1 in [4, 5, 6]] : List ℕ
```

The `x` should refer to the *second* `x`, not the first.

byorgey:
This shouldn't have much (if any) effect, but `unbound-generics` is a more modern reimplementation of `unbound` based on the support for generic types in GHC, instead of on the generics library `RepLib`.  In the long term, there is a much better chance the former will continue to be supported than the latter.  (I was kind of hoping switching might also fix #44 but no such luck.)

The only things to be aware of are:

1. `translate` is now `coerce`
2. When you add a new data type you may need to add `deriving Generic` to it.

### DIFF
--- a/disco.cabal
+++ b/disco.cabal
@@ -64,7 +64,7 @@ library
                        split >= 0.2 && < 0.3,
                        transformers >= 0.4 && < 0.6,
                        containers >=0.5 && <0.6,
-                       unbound >= 0.4 && < 0.6,
+                       unbound-generics >= 0.3 && < 0.4,
                        lens >= 4.14 && < 4.16,
                        exact-combinatorics >= 0.2 && < 0.3,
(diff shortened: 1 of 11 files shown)


callahanhirrel:
Is it safe to assume `deriving Generic` will be a part of all data types now? If not, when will it not be used?

byorgey:
We only need it for data types which will be part of some kind of syntax tree.  With that said, it doesn't matter much either way: adding `deriving Generic` even if it is not ultimately needed doesn't really hurt, and if you forget it where it is needed, GHC will give an error like "Can't infer `Generic YourType`..."

PR #104, opened 2018-06-09
Fix multiplication parsing with literals in parens

### LINKED ISSUE #96: Function Application mixup with Multiplication
The example (4)(4) evaluates as follows:

```
Disco> (4)(4)
NotFun (ATNat TyN 4)
Disco> 4(4)
16
```
Both of the above examples should return 16, but expression `(4)(4)` is parsing incorrectly.

SanjitKal:
Changed isMultiplicativeTerm function in Parsing file to recursively check if term inside TParens is multiplicative


SanjitKal:
I still need to add a few tests which emphasize this fix.
Ex: (4)(4)
      (4) 4

byorgey:
Great!  In addition to some tests could you also tweak the tutorial, if necessary, to make sure this is described accurately there?

byorgey:
OK, thanks for checking.

PR #107, opened 2018-06-28
Implemented the divides operation for finite types

### LINKED ISSUE #106: Divisibility testing is incorrect for Zn
For example:
```
Disco> (2 : Z7) divides (5 : Z7)
false
```
This should be `true`, since `(2 * 6 : Z7) = (5 : Z7)`.

SanjitKal:
Fixes #106.

This pull request fixes the issue with finite types divisibility testing by adding a special core constructor for divisibility with finite types. The divides operation is desugared to this new constructor if the operands are of a finite type. The finite types divisibility algorithm works as follows:
When checking if 'a' divides 'b' in a mod space 'm', 

1.) Let d = gcd(a,m)
2.) If d divides b, return true. Otherwise, return false.


src/Disco/Interpret/Core.hs
```diff
@@ -556,6 +559,15 @@ modDiv n [c1,c2] = do
     Nothing -> throwError DivByZero
 modDiv _ _ = error "Impossible! Wrong # of cores in modDiv"
 
+modDivides :: Integer -> [Core] -> Disco IErr Value
+modDivides n [c1,c2] = do
+  VNum _ a <- whnf c1
+  VNum _ b <- whnf c2
+  return $ mkEnum $ divides (toRational (binaryGCD (numerator a) n)) b
```
byorgey: The [documentation for the `binaryGCD` function](https://hackage.haskell.org/package/arithmoi-0.7.0.0/docs/Math-NumberTheory-GCD.html) states that it is very slow for `Integer`, which is the type we are using it at here.  Let's just use the standard `gcd` function instead.

byorgey: Also, does this work when `a`, `b`, or both are zero?  Can we add some more tests to cover those cases?

Issue #119, opened 2018-07-14
Prims


byorgey:
The idea is to introduce a new syntax for "primitives", `$prim`, which are parsed as a special `TPrim` node in the AST and are handled specially.  A great first example to do in this style is `isPrime`.  We should be able to write something like this:
```
isPrime : N -> Bool
isPrime = $isPrime
```
So we declare a name `isPrime` along with its type, and simply define it in terms of a prim (also called `isPrime`).  This way, we don't need special typechecking or parsing rules for `isPrime` at all.  We just need to translate it into a `Core` prim which gets interpreted by calling an efficient primality testing function in Haskell.

As for typechecking, we might still want to have a typechecking rule for `isPrime` in order to prevent people from writing bogus things like
```
foozle : List Z -> Bool -> N * Q
foozle = $isPrime
```
but perhaps a different approach would be to add a flag enabling the use of prims, which is off by default.  So if you turn on the flag (as we would in e.g. the standard library) you are acknowledging that you have to use the prims responsibly, or else the resulting program might crash.

PR #120, opened 2018-07-16
Desugar

### LINKED ISSUE #66: Decouple desugaring and compilation
The `Disco.Desugar` module is really mixing two different phases: *desugaring* and *compilation*.  We should split this apart: one phase to desugar `ATerm -> ATerm` and another phase to compile `ATerm` to `Core`.

Note, we will want to do #61 first, so when desugaring the `ATerm` we get something that is statically known to have fewer constructors and so on.

### LINKED ISSUE #67: Lazy short-circuiting 'and' does not work with chains
```
Disco> 2 < 1 < 1/0
DivByZero
```
This should be `false`, because it desugars to  `2 < 1 and 1 < 1/0` which in turn should desugar to a lazy case analysis.

### LINKED ISSUE #68: Remove OAnd, OOr
These are not needed in the core language anymore, after fixing #65 by desugaring.

### LINKED ISSUE #98: Unnecessary lambdas when desugaring and, or?
Right now we desugar something like `t1 and t2` to a lambda with a case applied to `t1` and `t2`.  It seems like we could simply desugar to the case expression with `t1` and `t2` substituted inside it.

byorgey:
Split what used to be called "desugaring" into two phases: 

1. Desugaring (`Disco.Desugar`) now goes from `ATerm` to a new intermediate AST `DTerm`, which lives in `Disco.AST.Desugared`.  `DTerm`s are still type-annotated like `ATerm`s, but are much simpler and have fewer features (for example, no comprehensions, no lets, lambdas always have a single argument, no boolean guards, *etc.*). The new AST type was super easy to construct thanks to the new Trees That Grow framework. =)

2. Compilation (`Disco.Compile`) goes from `DTerm` to `Core`.  It doesn't do a whole lot except erase types, translate all constructors into `CCons`, and deal with some overloading (*e.g.* distinguish operators that have to compile to different runtime operations depending on their type).

I also added some additional desugaring, *e.g.* got rid of a whole bunch of core operations that we can just desugar instead, as well as desugaring list comprehensions to monadic operations.  Right now the list comprehension desugaring requires you to have things called `concat` and `mapList` in scope (and it would fail if they have the wrong type).  It is also much slower than it used to be.  Once we add prims (#119) we should have list comprehensions instead desugar to the use of a few prims representing these operations which can then be interpreted using the interpreter code we were using before (`vmap`, `vconcat`, *etc.*).  We will need to desugar other container comprehensions similarly (using prims) anyway. 

Along the way I also fixed a few bugs, added a bunch of comments, etc.  I also reorganized `AST.Generic` and moved all the `Alpha` and `Subst` instances from the specific AST modules (`Surface`, `Typed`) to polymorphic instances in `AST.Generic`, so no new instances need to be written when creating a new AST variant.

One thing that could still be improved is that `DTerm`s still contain `BOp` or `UOp` values, even though only some of them can possibly occur (because the other ones get desugared).  It would be nice to know statically which operators can occur in which ASTs.  We should also eventually write a type checker for `DTerm` --- writing such a typechecker is very easy since `DTerm` values are explicitly annotated with types, and it can serve as a good sanity check on the desugaring process.

Take a look and see what you think.  Happy to hear your questions/comments/suggestions.  I'll let you merge it once you've had a chance to look at it and think it's ready.

Fixes #68. Fixes #67. Fixes #66. Fixes #98.


SanjitKal:
I think these changes look good, so I'm going to merge them.

PR #122, opened 2018-07-24
Type definitions

SanjitKal:
This branch introduces recursive type definitions into the Disco language. I resolved merge conflicts with master. Let me know if this is good to go/if there needs to be other work done.


docs/tutorial/typedefs.rst
```diff
@@ -0,0 +1,77 @@
...
+****************
+Type Definitions
+****************
+
+Disco supports recursive type definitions. Let's look at a few example.
+
+Consider the following function which takes a list of natural number triplets and
+returns the sum of all the triplets in the list:
+
+::
```
byorgey: Code blocks such as this one should be taken from the actual `tydefs.disco` file using the `literalinclude` directive.  You can tell it to display only a portion of the file.  See the documentation here: http://www.sphinx-doc.org/en/1.5.1/markup/code.html .  Doing it that way means that the code that is displayed comes directly from a file which gets checked every time the test suite runs; as it is there is no guarantee that the code in the tutorial has been kept up to date with the code in `examples/tydefs.disco`.

byorgey: See other files in the tutorial for examples of `literalinclude` (though I don't know if there are any examples of including only a portion of the file).

src/Disco/Parser.hs
```diff
@@ -352,6 +355,12 @@ parseDefn = label "definition" $
   <$> ident
   <*> (indented $ (:[]) <$> (bind <$> many parseAtomicPattern <*> (symbol "=" *> parseTerm)))
 
+-- | Parse the definition of a user-defined algebraic data type.
+parseTyDefn :: Parser Decl
+parseTyDefn = label "ADT defintion" $
```
byorgey: This should say "type definition" instead of "ADT definition"

src/Disco/Parser.hs
```diff
@@ -697,13 +706,18 @@ parseAtomicType = label "type" $
...
   <|> TyVar <$> parseTyVar
   <|> parens parseType
 
 parseTyFin :: Parser Type
 parseTyFin = TyFin  <$> (reserved "Fin" *> natural)
          <|> TyFin  <$> (lexeme (string "Z" <|> string "ℤ") *> natural)
 
+-- | Need to change identifier to take an extra parser argument
+parseTyDef :: Parser String
+parseTyDef =  ((:) <$> upperChar <*> identifier)
```
byorgey: It looks like this isn't finished yet (it won't allow single-character names).  I forget what we discussed---something about modifying `identifier` to take a parser as an argument to specify what kind of first character to accept?

src/Disco/Parser.hs
```diff
@@ -714,12 +714,11 @@ parseTyFin :: Parser Type
...
          <|> TyFin  <$> (lexeme (string "Z" <|> string "ℤ") *> natural)
 
--- | Need to change identifier to take an extra parser argument
 parseTyDef :: Parser String
-parseTyDef =  ((:) <$> upperChar <*> identifier)
+parseTyDef =  identifier upperChar
 
 parseTyVar :: Parser (Name Type)
-parseTyVar = string2Name <$> identifier
+parseTyVar = string2Name <$> (identifier letterChar)
```
byorgey: Shouldn't this be `lowerChar` instead of `letterChar`?

SanjitKal: Oh, you're right, thanks.

byorgey:
I get this error:
```
Disco> takeStream 10 (streamy2 nats)
Impossible! No matching case in prettyWHNF for VCons 1 [VIndir 24]: TyDef "Lst"
CallStack (from HasCallStack):
  error, called at src/Disco/Pretty.hs:374:21 in disco-0.1.0.0-C1ivNbdZ4bODANbXRSSWkf:Disco.Pretty
```
I think the problem is that `prettyWHNF` doesn't know what to do with `TyDef "Lst"`; it should simply look up its definition and expand it.

PR #128, opened 2018-07-31
Strings and chars

### LINKED ISSUE #62: Primitive Character type for Unicode code points
It's not all that mathematical.  But when talking about *e.g.* encryption, or just for general usefulness of writing example programs that can produce messages, it might be nice to have very basic built-in support for handling strings.  Something very basic should be enough:

* A new primitive type `Character` which represents a Unicode codepoint.
* Built-in functions to convert between `Character <-> Integer`; unsure what their names should be.
* Special support for literal strings, which desugar to lists of characters.

There are some issues to be worked out regarding how lists of characters get pretty-printed; see #63.

SanjitKal:
Fixes #62.

This pull request adds characters and strings to the Disco language.


byorgey:
@SanjitKal You probably noticed that I turned on this `restyled` thing which automatically runs `stylish-haskell` on any PRs.  If it notices any changes it opens a new PR (against your branch, not master) with the formatting changes.  When you see those, if you're happy with the formatting changes you can go ahead and merge the PR it generated (in this case #129).

test/types-tydefs/expected
```diff
@@ -1,7 +0,0 @@
-Loading test/types-tydefs/types-tydefs.disco...
```
byorgey: Was there a reason to delete these files, or was it just a mistake?

SanjitKal: This was a mistake, thanks.

byorgey:
Looks good. I think the only thing left is to add some more tests reflecting the additional issues that have been pointed out and fixed.

byorgey:
We'll also want primitive functions to convert between Char <-> Nat, but that can wait for later.

PR #132, opened 2018-08-03
Modules

SanjitKal:
This pull request introduces a module system to Disco.


src/Disco/Eval.hs
```diff
@@ -186,6 +189,18 @@ withEnv = local . const
 -- | Errors that can be generated during interpreting.
 data IErr where
 
+  -- | Module not found.
+  ModuleNotFound :: ModName -> IErr
+
+  -- | Cyclic import encountered.
+  CyclicImport :: ModName -> IErr
+
+  -- | Error encountered during typechecking XXX. Is this okay?
```
byorgey: You can get rid of these `XXX`s, as we discussed adding these to `IErr` is great.

src/Disco/Parser.hs
```diff
@@ -262,11 +263,13 @@ wholeModule = between sc eof parseModule
...
 parseModule :: Parser Module
 parseModule = do
+  imports <- many parseImport
   topLevel <- many parseTopLevel
-  let theMod = mkModule topLevel
+  let theMod = mkModule imports topLevel
   return theMod
   where
-    groupTLs :: [DocThing] -> [TopLevel] -> [(Decl, Maybe (Name Term, [DocThing]))]
+    -- XXX: Change type of groupTLs to be tuple ([ModName],[(Decl, Maybe (Name Term, [DocThing]))])
```
byorgey: Delete this comment

src/Disco/Parser.hs
```diff
@@ -300,6 +303,10 @@ parseTopLevel = L.nonIndented sc $
       TLDoc  <$> parseDocThing
   <|> TLDecl <$> parseDecl
 
+parseImport :: Parser ModName
+parseImport = L.nonIndented sc $
+      reserved "import" *> identifier letterChar
```
byorgey: `identifier letterChar` won't cut it if we want to be able to write things like `import foo/bar/baz.disco`.  I guess we'll want something like  ``many1 filenameChar `sepBy` (symbol "/") <* option (string ".disco")`` but I'm not convinced of the exact details.  Also we'll have to define `filenameChar` to say exactly what characters are allowed in a file or directory name.

src/Disco/Typecheck.hs
```diff
@@ -85,30 +85,37 @@ inferTelescope inferOne tel = do
 
 -- | Check all the types in a module, returning a context of types for
```
byorgey: We should update this comment.  It looks like it was already out of date *before* we made these changes.

src/Disco/Typecheck/Monad.hs
```diff
@@ -58,24 +59,41 @@ instance Subst Type Defn
...
-  { _termDefns :: Ctx ATerm Defn
-  , _tyDefns   :: TyDefCtx
+-- | A typing context is a mapping from term names to types.
+type TyCtx = Ctx Term Sigma
+
+-- | Type checking a module yields a value of type ModuleInfo which contains
+--   mapping from terms to their relavent documenation, a mapping from terms to
+--   properties, and a mapping from terms to their types.
+data ModuleInfo = ModuleInfo
+  { _docs     :: Ctx Term Docs
```
byorgey: We should add the like `makeLenses ''ModuleInfo` after the definition of `ModuleInfo`.  But we should also change the names of the fields --- as it stands the names are so short and generic that they are likely to conflict with lots of local variables.  Let's change them to *e.g.*  `_modDocs`, `_modProps`, and so on.

src/Disco/Typecheck/Monad.hs
```diff
@@ -58,24 +59,41 @@ instance Subst Type Defn
...
+emptyModuleInfo :: ModuleInfo
+emptyModuleInfo = ModuleInfo emptyCtx emptyCtx emptyCtx M.empty emptyCtx
 
--- | The initial, empty @DefnCtx@.
-emptyDefnCtx :: DefnCtx
-emptyDefnCtx = DefnCtx
-  { _termDefns = emptyCtx
-  , _tyDefns   = M.empty
-  }
+-- | To combine two values of type ModuleInfo, we first make sure that there
```
byorgey: Delete this old commented-out code.

src/Disco/Typecheck/Monad.hs
```diff
@@ -112,7 +130,7 @@ instance Monoid TCError where
 --   variables and their types and a read-write context of term and
 --   type definitions; collects constraints; can throw @TCError@s; and
```
byorgey: This comment should be updated.

byorgey:
Were you in the middle of working on any other commits to this branch (or planning to)?  If not I can just fix the test and merge; I'd rather just get it merged and file the remaining issues as tickets.

(In other news arithmetic patterns are just about done!)

SanjitKal:
Sorry, I didn’t get around to working on the remaining fixes yet. You can
go ahead and fix the test and merge.

That’s great! Excited to try them out soon:)

PR #134, opened 2018-08-07
Pattern/guard refactoring, and arithmetic patterns

### LINKED ISSUE #11: Arithmetic patterns
In mathematics it is common to define functions by pattern-matching on arithmetic expressions, for example:
```
f 0 = 1
f (2k+1) = f k
f (2k+2) = f k + f (k+1)
```
This is unambiguous and well-defined as long as the arithmetic expression represents an invertible function of a single variable.  For example, `f (x+y) = ...` should not be allowed since it has multiple variables (specifically, because it does not uniquely determine `x` and `y`); `f (0*x)` should not be allowed because `x -> 0*x` is not invertible, so matching it against an input does not uniquely determine `x`.

I made a first attempt at this in the `arithpat` branch, and found that it is surprisingly tricky to get all the details right.  But perhaps I was just trying to be too general.  A simple first pass might be to allow only multiplication and addition and matching only on natural numbers.

We will also want to allow matching on negations of things---this is how one can define functions by pattern-matching on negative integers (which is currently not possible!).

However, we probably also want to eventually allow things like `f (p/q) = ...` for matching on rational numbers; in theory this is ambiguous, but by convention it makes sense for this to match on the numerator and denominator.  The question is what `f (a/2) = ...` should mean.  What if `f` is called on `3/4`?  Should it (a) fail to match because the denominator is not `4`, or (b) match and bind `a` to `3/2`?  I kind of lean towards (a) even though it seems like more of a hack mathematically.  I suppose the idea would be that when pattern-matching on rational numbers we think of them as ordered pairs rather than as arithmetic things.

byorgey:
I added arithmetic patterns such as  `f(2x + 1) = ...` or  `f(-p/q) = ...`.  Along the way I also did a bunch of refactoring.

- We now parse patterns by parsing a *term* and then running the function `termToPattern` to try to convert it to a pattern.  This makes the parsing much more sane than it otherwise would have been when adding arithmetic patterns; it also will make it easier to do things like add pattern binders in comprehensions later.  Patterns are almost a subset of terms, except for wildcards; so there is a new `TWild` constructor for terms that gets parsed, but rejected at typechecking time if you try to use a wildcard in an actual term.

- Got rid of "successor" patterns (like `S n`) since those can now be written `n+1`.

- I added better pretty-printing for desugared terms since it was helpful for debugging, by erasing `DTerms` to `Surface` terms and then pretty-printing.

- Refactored the desugaring of function definitions so it now generates an `ATerm` case expression and recursively desugars it, instead of directly generating a `DTerm` as before.  This means there is now only one place in the code where we have to handle desugaring pattern matching.

- Case expressions now support a new kind of `let` guard, so you can write *e.g.*
    ```
    {? x if 2 divides z let x = z // 2 if x > 5, 900 otherwise ?}
    ```
    I thought I needed these for desugaring arithmetic patterns, though it turns out not to be strictly necessary since `let x = t` just desugars to `when t is x`.  But they seemed generally useful, and easy to implement, so why not?

- Guards can now desugar to *multiple* guards (see the new `desugarMatch` function).  This is necessary for desugaring arithmetic patterns, but it turns out to be helpful for other things as well.  In particular, I was having trouble getting lazy matching to work properly until I realized that patterns in the desugared language should contain *only variables*, so nested patterns have to desugar to a sequence of guards doing the matching piece by piece.  This makes the order of matching explicit and allows the matching to quit early *without* needing any special support in the interpreter: now all it has to do is reduce a value to WHNF, match it against a pattern, and bind the resulting subvalues to variables.

Closes #11.


byorgey:
Amazingly, this PR has no merge conflicts with the modules branch we just merged!  I think this means (1) we are doing a good job organizing code so unrelated stuff is in separate places, (2) the restyled bot is helping cut down on spurious conflicts due to formatting, or maybe (3) we just got lucky.

Issue #195, opened 2020-06-12
'lg(2^35 - 1)' example results in 'Unsolvable NoUnify'


enlambdment:
In the [Arithmetic > Advanced Arithmetic](https://disco-lang.readthedocs.io/en/latest/tutorial/arithmetic.html#advanced-arithmetic) section of the disco docs, there is an example:

```
Disco> lg (2^35 - 1)
34
```

However, I currently get `Unsolvable NoUnify` when I run the same example directly in the disco REPL.

byorgey:
Thanks for reporting!  I think what's going on is that `lg` has type `N -> N` but `2^35 - 1` can't have type `N` because it uses subtraction.  This is something I've gone back and forth on a lot, and at the time the tutorial was written, apparently `2^35 - 1` could still be given the type `N`.  But since then I've come down on the side of saying that anything using subtraction must have type `Z` or `Q`, and also added a new saturating subtraction operator `.-` which has type `N -> N -> N`.  So all that would be needed to fix the example is to replace `-` by `.-`.

enlambdment:
Thanks for clarifying. Looking at this now.
I will change both examples `sqrt (299^2 - 1)`, `lg (2^35 - 1)` to use the saturating subtraction operator.
(Please let me know if you prefer a 2nd issue opened for the separate edit involving `sqrt` example.)

byorgey:
No, one issue for both is just fine!

byorgey:
Closed by #196 .

PR #196, opened 2020-06-12
Revising sqrt, lg examples to use saturating subtraction

enlambdment:
This is a revision to the docs tutorial to address issue #195 . 
Examples using `sqrt`, `lg` now use saturating subtraction because these operations are not defined over negative numbers in disco.


byorgey:
Thanks!

Issue #182, opened 2020-02-22
Allow patterns in lambda bindings?


byorgey:
For example, `(λ(x,y). x + y) : N*N -> N`?  It's probably not worth having syntax for multiple clauses though.  So this would only be useful for things like tuples.

byorgey:
This is much more important now that things are uncurried.  So for example if you want to do a fold with a non-standard operator you would want to write something like
```
reduce(\(x,y). x+2y, 0, bs)
```
The alternative is to define a separate named helper function, or write an inline case, neither of which is very nice:
```
f (x,y) = x+2y
reduce(f, 0, bs)

reduce(\p. {? x+2y when p is (x,y) ?}, 0, bs)
```

byorgey:
This requires
* Updating AST to allow patterns in lambda binders instead of just names (but only in surface and typed syntax)
* Updating parser to parse it
* Updating the typechecker
* Updating the desugarer to desugar pattern bindings into normal bindings + case.  In fact, we already do this for definitions (the difference being that with definitions there can be multiple clauses).  Should find a way to reuse the code.

byorgey:
Closed by #198.

PR #198, opened 2020-06-13
Lambda Patterns

shaylew:
Allow patterns (rather than just variables) as lambda arguments. See #182.

What this patch implements is fairly permissive. In addition to the expected simple cases, like
```
let swap = \(x, y). (y, x) in swap ('a', 'b')
```
it also allows some somewhat more questionable constructs like
```
let fromLeft = \left x. x in fromLeft right "uh-oh"
let almostNowhere = \1 . True in almostNowhere 0
```
That said, there's no exhaustivity checking anywhere else and this doesn't seem like the right place to wade into adding it.

The fact that we now accept type ascriptions in patterns also means we accept them in *definition* patterns, which is probably harmless but can allow some odd definitions. For instance: 
```
weird : N -> Q
weird 0 = 0
weird (1 : Z) = 1
weird (x : F) = x
```
The pattern is checked against the declared type (so if we had `weird : Z -> Q` the last clause wouldn't be able to match on `x : F`) and it does in fact bind its variable at the type ascribed to it (so if we had `weird : N -> Z` the last clause wouldn't be able to return `x` after binding it at type `F`), so nothing is truly amiss here even if it looks a little funny.


src/Disco/AST/Surface.hs
```diff
@@ -373,6 +376,11 @@ pattern PVar name = PVar_ () name
 pattern PWild :: Pattern
 pattern PWild = PWild_ ()
 
+ -- (?) TAscr uses a PolyType, but without higher rank types
+ -- I think we can't possibly need that here.
```
byorgey, line 380: Agreed.

src/Disco/Typecheck.hs
```diff
@@ -1091,6 +1084,14 @@ checkPattern (PVar x) ty = return (singleCtx x (toPolyType ty), APVar ty (coerce
 
 checkPattern PWild    ty = return (emptyCtx, APWild ty)
 
+checkPattern (PAscr p ty1) ty2 = do
+  -- We have a pattern that promises to match ty1 and someone is asking
+  -- us if it can also match ty2. So we just have to ensure what we're
+  -- being asked for is a subtype of what we can promise to cover...
```
byorgey, line 1090: Thanks for writing this comment, it's tricky to think about getting this check the right way around, but I agree this is correct.  In retrospect it makes sense that subtyping for type ascriptions in patterns would be the opposite way around from type ascriptions in expressions.

test/props-fail/expected
```diff
@@ -1,10 +1,10 @@
 Loading bad-tests.disco...
 Running tests...
   badmap:
-  - Test failed: badmap (λ(x : ℚ). x / 0) [3, 4, 5] == [6, 7, 8]
+  - Test failed: badmap (λx. x / 0) [3, 4, 5] == [6, 7, 8]
```
byorgey, line 4: Was this change related to erasure?

byorgey:
Thanks!  This is great.  I'm really pleased that this ended up actually *simplifying* the code in many places!

Sorry about the CI failure, that was my fault.  I checked out this PR and ran the test suite locally just to make sure it's OK.

test/props-fail/expected
shaylew: Yeah. With pattern type ascriptions being their own thing now it seemed more convincing that they should erase the way expression type ascriptions do.

byorgey:
Makes sense.

PR #200, opened 2020-06-14
WIP: props

shaylew:
The good (things that work):

- Quantified `Prop`s typecheck, also at type `Prop`.
- `holds` can search some types successfully. Currently that includes finite types (including finite function spaces), all numeric types, and all sums and products of searchable types. It does not yet include infinite function spaces or containers. 

The bad (things that don't work):

- Curried and uncurried quantifications behave differently, because right now product types are searched fairly (i.e. interleaving both sides) but multiple layers of quantifiers are searched one argument at a time. Probably the sensible thing to do is to gather up the layers in the compiler (Core can already represent that) and search all arguments together.

- All props print as `<Prop>` right now.

The ugly (implementation notes):

- In the interpreter a `Prop` is represented as a `VCons`, with four possible constructors. I'm not sure how much in the spirit of how the rest of the interpreter works this all is, but it seemed sensible at the time.

  - `VCons 0 counterexamples`: false, and the argument is a list of counterexamples found to a prefix of universal quantifiers.
  - `VCons 1 witnesses`: true, and the argument is a list of witnesses found for a prefix of existential quantifiers.
  - `VCons 2 [ty, f]`: universal quantification, where `f` accepts one argument of type `ty` and returns a prop.
  - `VCons 3 [ty, f]`: existential quantification, where `f` accepts one argument of type `ty` and returns a prop.

  There isn't currently any way to see the witness returned by search, but I imagine prop-based tests will want to use it.

- I wrote a separate version of enumerate (enumerateInf) that works on infinite types. The reason it has to be separate is we can't both return results in sorted order and guarantee we get to every value eventually (e.g. the standard order on pairs is lexicographic, which isn't well-founded). It also doesn't currently work for all types. 

  (It's starting to look reasonable to pull in some sort of general enumeration library, if we want to push the set of enumerable types much past this point, but I haven't done the research on what's out there.)

- Because not all types are searchable, we're going to end up having to pick one of these ways to deal with it:

  1. Emit an error when someone calls `holds` on a non-searchable prop. (This is the current behavior.)
  2. Make it a type error to quantify over a non-searchable type in forall and exists.
  3. Split out a `SearchableProp` subtype of `Prop`, and have the type of `forall (x : T). p x` depend on whether T is searchable. 

  I'd lean towards (2) since right now the only thing you can do with a prop is search it, but if you have other uses in mind then other options might appeal.


byorgey:
Thanks!  I may or may not have time to look at this today, then I will be out of town for a couple days, but will definitely be able to offer some feedback later in the week.

byorgey:
You may find this interesting: https://github.com/disco-lang/disco/tree/enumeration
I honestly don't remember what's there, and I seem to recall that it ended up being much too complex, but it might have some interesting ideas.

byorgey:
> Probably the sensible thing to do is to gather up the layers in the compiler (Core can already represent that) and search all arguments together.

That makes sense to me.

byorgey:
> Make it a type error to quantify over a non-searchable type in forall and exists.

I also think this is the best option.  Indeed, there's not much you can do with a `Prop` that quantifies over a non-searchable type.  I would be perfectly happy to get something merged that only allows quantifying over a manageable subset of types (even if it would be theoretically possible to support more) and then perhaps add more supported types later.

shaylew:
Made some progress on tighter typechecking for quantifiers (and for enumerate, which seemed to be right next door). I just added `QFinite` and `QSearch` qualifiers and corresponding entries in `hasQual` and `qualRulesMap`.

... But it's not clear how to extend this approach to handle uninhabited types correctly. It's not quite the case that a product is only finite (or searchable) if both factors are, since `Void * (N -> N)` is perfectly tractable; but the way qualifier (and sort) rules work it seems like they can't express "either one of the factors must be empty, or else they both must be searchable". This is perhaps more of a pain for enumerate (which was working before, albeit with some strange results for types that weren't enumerable), but it'd be nice to have it working. 

What's frustrating is how easy this is to answer for ground types -- it's just a matter of teaching the constraint solver to puzzle through it incrementally somehow.

Speaking of which, I haven't dug into the constraint solver to figure out what the deal is, but I've run into something strange:

```
Disco> :type ∃ a. a + a == a
Unsolvable NoUnify
Disco> :type \ a. a + a == a
λa. a + a == a : ℕ → Bool
```

Lambda and quantifier abstraction is handled by the same code with a few detours, and it seems like the culprit is this bit since with it commented out the quantifier abstraction defaults the type to `N` the same as the lambda.

https://github.com/disco-lang/disco/blob/d2ff3ce62987bf33650a1589492b3d6103a2dcfa/src/Disco/Typecheck.hs#L871-L874

I'm not really sure how to peek at what the solver thinks is going on here, and I'm not sure why putting another constraint on the argument type here breaks things. There only seems to be a problem when (1) we're doing arithmetic on the argument (`forall a. a <= a` is fine) and (2) no literals are involved (`forall a. a + a == 0` is also fine). So it does look like there's only an issue when defaulting happens.

byorgey:
>  just added `QFinite` and `QSearch` qualifiers and corresponding entries in `hasQual` and `qualRulesMap`.

Yeah, as you discovered, I don't know if this is going to work, because the logic for deciding this doesn't seem to fit into the framework for type qualifiers.  But does it need to?  Will there ever be a situation where we need to decide finiteness or searchability for a type with type variables in it?  I'm honestly not sure.

shaylew:
Yeah, that's a pretty convincing reason to not do defaulting there! Requiring an annotation makes sense, and could be relaxed in qualified types mode if that ends up being a thing. (It's pretty much the same as haskell's defaulting/ambiguous type variable situation, where there are some kind of defaulting you want at the repl for usability, some places where a type variable really is irrelevant so you can zonk it to (), and some places where runtime behavior depends on the type you choose so you have to ask the user to fix it.)

In addition to unsolved (possibly ambiguous) unification variables ending up being quantified over, it's also possible for skolems introduced by type signatures to end up in quantifiers.
```
reflexive : (a -> a -> Prop) -> Prop
reflexive p = forall x. p x x
```
This is also something we can't allow without a story for qualified types involving either type passing or evidence passing, because there's no way to compile this so it knows what type to search over. Luckily, there's also no way to write a type signature for `x` here, so requiring explicit types disallows it.

It also takes care of this last example, which is perfectly searchable (no matter what `a` is, and without even needing to know `a`), but which the sketched-out qualifiers framework wasn't able to deal with. 
```
vacuous : (a * Void -> Prop) -> Prop
vacuous p = forall x. p x
```

shaylew:
Here we go! I put in the fix to gather up nested abstractions so `holds` can search them together (fairly), and simplified the searchability checking. This version just checks that the type of the pattern is definitely searchable in isolation, before the solver gets involved.

byorgey:
BTW, re: a generic enumeration library, we can use http://hackage.haskell.org/package/simple-enumeration which I wrote for this exact purpose. I had just forgotten about it. =)

byorgey:
I tried `holds (exists (x:N). exists (y:N). exists (z:N). x > 0 && y > 0 && z > 0 && (x^2 + y^2 == z^2)` which I expected to yield `true` fairly quickly but it just hung for a long time.  How would the enumeration work in that case?

byorgey:
Wow, it's amazing how much code we get to delete when switching to use `simple-enumeration`.  I see that you omitted a case for `TyBag` from `enumerateType`, which is perfectly OK --- as I recall, bags turned out to be the big sticking point when I was working on this before.  I can't remember exactly why, but I remember it being a very thorny issue.  In retrospect, I think I let my stubbornness and perfectionism get the better of me --- the right solution (for now at least) is obviously to just throw up our hands and say "nope, can't enumerate bags, sorry"!

byorgey:
Gathering nested abstractions definitely doesn't seem to be working:
```
Disco> :compile (exists (x:N). exists (y:N). exists (z:N). x^2 + y^2 == z^2)
CCons 3 [CAbs (<[arg0]> CCase [<TelCons (<<({CVar 0@0},CPVar x1)>> TelEmpty)> CCons 3 [CAbs (<[arg02]> CCase [<TelCons (<<({CVar 0@0},CPVar y3)>> TelEmpty)> CCons 3 [CAbs (<[arg04]> CCase [<TelCons (<<({CVar 0@0},CPVar z5)>> TelEmpty)> CApp (CConst (OEq (TyAtom (ABase N)))) [(Lazy,CCons 0 [CApp (CConst OAdd) [(Lazy,CCons 0 [CApp (CConst OExp) [(Lazy,CCons 0 [CVar 4@0,CNum Fraction (2 % 1)])],CApp (CConst OExp) [(Lazy,CCons 0 [CVar 2@0,CNum Fraction (2 % 1)])]])],CApp (CConst OExp) [(Lazy,CCons 0 [CVar 0@0,CNum Fraction (2 % 1)])]])]]),CType (TyAtom (ABase N))]]),CType (TyAtom (ABase N))]]),CType (TyAtom (ABase N))]
```
Notice there are three nested `CCons 3` nodes but we only want one.   I'm going to see if I can figure out why.

byorgey:
Ah, it's because nested quantifiers desugar to alternating quantifiers and cases:
```
Disco> :desugar (exists (x:N). exists (y:N). exists (z:N). x^2 + y^2 == z^2)
∃arg0. {? ∃arg01. {? ∃arg02. {? x ^ 2 + y ^ 2 == z ^ 2 when arg02 is z 
                             ?} when arg01 is y 
                  ?} when arg0 is x 
       ?}
```
1. In this example the cases are all silly and redundant.  `desugarAbs` unconditionally puts in an `ATCase`; we should specially avoid generating a case expression in the common case that the pattern is a single variable or wildcard.
2. However, in general we'll need to deal with this, because we could have more complex patterns in quantifiers which actually require cases.  So maybe `unbindDeep` should look through case expressions somehow, if they only have a single branch.

byorgey:
By the way, I've given you push access to the `disco` repo.  I suggest you push this branch onto the `prop` branch in the `disco` repo, close this PR, and re-open a new PR from the `prop` branch to the `master` branch within the `disco` repo.  The benefits are that (1) integration with the restyling thing is much smoother, and (2) it's much easier for me (or others) to help since we can all push to the `prop` branch.  Of course we lose all the discussion here but we can easily link back to this closed PR.

Issue #189, opened 2020-06-05
Add OEIS interface


byorgey:
Using the `oeis` package.  Provide some OEIS primitives that allow interaction with the OEIS database.

metadave:
can you provide examples of what this could look like?

byorgey:
Sure!  The [oeis package](hackage.haskell.org/package/oeis) provides a way to retrieve all the info related to a sequence, but I am thinking of keeping things simple.  Maybe we could have two primitives: one to extend a sequence, and the other to yield the URL of a sequence.  So, in more detail:

* Add two new primitives, called something like `lookupSequence` and `extendSequence`.
    * Add them in [`Disco.Syntax.Prims`](https://github.com/disco-lang/disco/blob/master/src/Disco/Syntax/Prims.hs), in both the `Prim` data type and the `primTable`.  Probably make them non-exposed (i.e. set the boolean in the `primTable` to `False`).
    * Add them to [the `inferPrim` function in `Disco.Typecheck`](https://github.com/disco-lang/disco/blob/master/src/Disco/Typecheck.hs#L474).  I'm imagining type `[N] -> String` for `lookupSequence` and `[N] -> [N]` for `extendSequence`.
    * Add corresponding constants to the [`Op` data type in `Disco.AST.Core`](https://github.com/disco-lang/disco/blob/master/src/Disco/AST/Core.hs#L98), and update [the `compilePrim` function in `Disco.Compile`](https://github.com/disco-lang/disco/blob/master/src/Disco/Compile.hs#L120) to trivially turn the new prims into the new `Op` constants.
    * Update [the `whnfOp` function in `Disco.Interpret.Core`](https://github.com/disco-lang/disco/blob/master/src/Disco/Interpret/Core.hs#L804) to handle the new constants.  This is where you would actually write code to call the Haskell `oeis` package and convert back and forth to Disco runtime values.
* Add a new library module `oeis.disco` in `lib/`, and have it simply re-export the prims (you can see how this is done for `num.disco` as an example.
* Add some test cases to the test suite to exercise it.

byorgey:
As a further example of what this might look like when it's working:
```
Disco> import oeis
Disco> :type extendSequence
extendSequence : List ℕ → List ℕ
Disco> extendSequence [0,1,1,2,3,5]
[...]

PR #202, opened 2020-06-15
add lookupSequence, extendSequence for OEIS access

metadave:
- I'm not very happy about the `oeisLookup` implementation, suggestions for cleanup welcome!
- ~I'll pull in style changes from #203 into this branch after review~ 
  - changes pulled into branch

See also: https://github.com/disco-lang/disco/issues/189


byorgey:
Thanks!  Sorry it's taken me a while to look at it, I was away for a few days with my family.

lib/oeis.disco
```diff
@@ -0,0 +1,7 @@
+using Primitives
+
+lookupSequence : List N -> Unit + List Char
```
byorgey, line 3: It would be good to include some documentation and examples attached to these functions here, using `|||` (documentation) and `!!!` (doctest/quickcheck) syntax.

src/Disco/Interpret/Core.hs
```diff
@@ -905,6 +909,9 @@ whnfOp OUntil          = arity2 "until"     $ ellipsis . Until
 whnfOp OCrash          = arity1 "crash"     $ whnfV >=> primCrash
 whnfOp OId             = arity1 "id" $ whnfV
 
+whnfOp OExtendSeq      = arity1 "extendSequence" $ whnfV >=> oeisExtend
+whnfOp OLookupSeq      = arity1 "lookupSequence" $ whnfV >=> oeisLookup
+
```
byorgey, line 914: It's not really necessary to call `whnfV` (which reduces a value to weak head normal form) before calling `oeisExtend` or `oeisLookup` here: weak head normal form on a list only tells you if it is empty or a cons, and they both call `fromDiscoList` which evaluates the entire list anyway.

src/Disco/Interpret/Core.hs
```diff
@@ -1504,3 +1511,37 @@ primValOrd (VCons i []) (VCons j []) = compare i j
...
+    case lookupSequence hvs of
+      Just result -> parseResult result
+      Nothing     -> return leftUnit
+  where
+    parseResult r = do
+          let sequence = getCatalogNum $ catalogNums r
+          l <- toDiscoList $ toVal ("https://oeis.org/" ++ sequence)
+          return $ VCons 1 [l] -- right "https://oeis.org/foo"
+    fromVNum (VNum _ x) = fromIntegral $ numerator x
+    fromVNum v          = error $ "Impossible!  fromVNum on " ++ show v
```
byorgey, line 1533: Since `fromVNum` is duplicated in both `oeisLookup` and `oeisExtend`, why not pull it out into its own top-level definition?  Yes, it's extremely partial, but it's OK as long as we document it.

src/Disco/Interpret/Core.hs
```diff
@@ -1504,3 +1511,37 @@ primValOrd (VCons i []) (VCons j []) = compare i j
...
+    getCatalogNum (n:_) = n
+    toVal = map (\c -> vnum (toInteger (ord c) % 1))
+    leftUnit = VCons 0 [VCons 0 []]
+
+oeisExtend :: Value -> Disco IErr Value
+oeisExtend v = do
+    vs  <- fromDiscoList v
+    let xs = map fromVNum vs
+    let newseq = extendSequence xs
+    toDiscoList $ map (\i -> vnum (i % 1)) newseq
```
byorgey, line 1544: The lambda could also be written `vnum . (%1)`.  Personally I find that easier to read, though tastes differ.

test/lib-oeis/input
```diff
@@ -0,0 +1,6 @@
+:load lib/oeis.disco
```
byorgey, line 1: Since it's in the stdlib, it should work to say `import oeis` instead of `:load lib/oeis.disco`.

test/lib-oeis/input
```diff
@@ -0,0 +1,6 @@
...
+lookupSequence [1,1,2,3,5]
+lookupSequence []
+
+extendSequence [1,3,5,7]
+extendSequence []
```
byorgey, line 6: Maybe let's also add a test doing lookup and extend on a long and/or random sequence of numbers that isn't in the OEIS.

src/Disco/Interpret/Core.hs
```diff
@@ -1504,3 +1511,37 @@ primValOrd (VCons i []) (VCons j []) = compare i j
 primValOrd (VNum _ n1)  (VNum _ n2)  = compare n1 n2
 primValOrd v1           v2
   = error $ "primValOrd: impossible! (got " ++ show v1 ++ ", " ++ show v2 ++ ")"
+
+
+------------------------------------------------------------
+-- OEIS
+------------------------------------------------------------
+
+oeisLookup :: Value -> Disco IErr Value
```
byorgey, line 1520: You said you weren't happy with the `oeisLookup` implementation but it seems fine to me.  What were you not happy with?

metadave:
@byorgey all great suggestions, I'll push some changes shortly. Thanks for the review!

What are your thoughts on `extendSequence [foo]` returning `left [foo]` if the sequence isn't found/extendable, and `right someExtendedSequence` upon success?

src/Disco/Interpret/Core.hs
metadave: I think it's mostly that this is my first Haskell PR :-)

byorgey: OK, fair enough. =)

byorgey:
It's true that having `extendSequence :: List N -> (List N + List N)` would be more informative, since you could tell the difference between a sequence that's not in the OEIS vs a sequence that is there and just happens to have exactly the same number of terms as you started with.  However, that doesn't sound like a super useful distinction, and I think the usability would suffer.  I imagine in many (most?) cases `extendSequence` will be used in a situation where you already know the sequence *is* there (because you know what sequence you want, or you already did `lookupSequence`), so having to pattern-match on the result to get out the list you know you're going to get would be annoying.

metadave:
@byorgey pushed some changes (squashed)

byorgey:
Looks great, thanks!

PR #205, opened 2020-06-18
make applying VFun values work

### LINKED ISSUE #204: enumerate returns bogus functions
When you try to enumerate a (finite) function type, you get the right-shaped list but attempts to take out the functions and apply them fail.

```
Disco> map(\f. f True, enumerate (B -> B))
Impossible! funArity on VFun_ <fun>
Disco> enumerate (B -> B) == [\_ . False]
Impossible! funArity on VFun_ <fun>
```

This turns out to be a reasonable complaint, because what `enumerate` gives back is a list of `VFun`s and `VFun`s don't record an arity. It's not reasonable to expect the interpreter to be able to apply the `VFun` without knowing its arity, either.

It turns out that currently `VFun` is only ever used at arity 1, so the simplest thing to do seems to just make that explicit (and turn the list of values it gets as an argument into a single value). But if there's some forward-looking reason to make it more general it's equally easy to add an arity to the constructor.

shaylew:
Closes #204 and adds tests for the actual behavior of the enumerated functions.

### DIFF
--- a/src/Disco/Eval.hs
+++ b/src/Disco/Eval.hs
@@ -178,12 +178,12 @@ data Value where
 -- | A @ValFun@ is just a Haskell function @[Value] -> Value@.  It is a
 --   @newtype@ just so we can have a custom @Show@ instance for it and
 --   then derive a @Show@ instance for the rest of the @Value@ type.
-newtype ValFun = ValFun ([Value] -> Value)
+newtype ValFun = ValFun (Value -> Value)
 
 instance Show ValFun where
   show _ = "<fun>"
 
-pattern VFun :: ([Value] -> Value) -> Value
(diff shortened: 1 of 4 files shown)


src/Disco/Eval.hs
```diff
@@ -178,12 +178,12 @@ data Value where
 -- | A @ValFun@ is just a Haskell function @[Value] -> Value@.  It is a
```
byorgey, line 178: Should change the comment to match.

byorgey: You know what, I'm just going to merge and then fix this comment myself, seems silly to wait for you to fix it...

byorgey:
Looks great!

Issue #78, opened 2018-04-15
`:browse` command for disco REPL


byorgey:
It would be nice to have a command to show the types of all the names which are currently loaded.  In GHCI this is called `:browse`; I am not particularly attached to that name if we come up with a better one.

Issue #79, opened 2018-04-15
`:help` for disco REPL


byorgey:
Right now the `:help` command just prints `Help!` which is a bit silly.  At the very least it should print out the available commands like `:doc`, `:load`, `:desugar`, and so on.

byorgey:
Closed by https://github.com/disco-lang/disco/pull/230 .

PR #208, opened 2020-06-20
add :browse REPL command

metadave:
does `topCtx` have everything that should be shown with `:browse`?

```
Disco> :load example/prog.disco
Loading prog.disco...
Running tests...
  fib: OK
Loaded.
Disco> :browse
f : (ℕ → ℕ) → ℕ × ℕ → ℕ → ℤ
fact : ℕ → ℕ
fact2 : ℕ → ℕ
fib : ℕ → ℕ
isEven : ℕ → Bool
isOdd : ℕ → Bool
```

I'll add tests once I'm sure the implementation is correct.

closes https://github.com/disco-lang/disco/issues/78

### DIFF
--- a/src/Disco/Interactive/Eval.hs
+++ b/src/Disco/Interactive/Eval.hs
@@ -100,6 +100,7 @@ handleCMD s = do
     handleLine (Doc x)       = handleDocs x
     handleLine Nop           = return ()
     handleLine Help          = iputStrLn "Help!"
+    handleLine Names         = handleNames
 
 handleLet :: Name Term -> Term -> Disco IErr ()
 handleLet x t = do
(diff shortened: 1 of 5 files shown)


byorgey:
There's no specification of what `:browse` ought to show!  I think showing everything in `topCtx` makes sense.

Also, I said `:browse` just because that's what the command is called in `ghci`, but I'm not convinced it's the best/most obvious name for students.  Bikeshedding welcome.

metadave:
I'll suggest `dir` from Python: https://docs.python.org/3/library/functions.html#dir

byorgey:
Hmm, I'm not convinced `dir` is any better than `browse`.  What does `dir` mean?  Would it have any particular meaning/connotation for a student? I suppose it stands for "directory" but that it's only meaningful if you have experience with DOS.  Also not all students will know python and even those that do might not be familiar with `dir` (I wasn't!).

Maybe something like `:list`... hmm, no, that's already a keyword built into the language.  `:context` might work?  I don't know if it's too technical.

metadave:
What if this functionality was just built into `:help` (#79)?

byorgey:
`:names` could be good. Or hmm, how about something like `:catalog`? Or we could make it like a text adventure game and call it `:inventory`.  Or `:showmeallthethings` ;-) .

I think of this command and `:help` as serving very different roles.  `:help` means "I don't know how to use this REPL thingy, what can I do?"  Whereas `:browse`, whatever it is called, means "I want to see the names of all the things currently imported".  Those are very different situations and I think it would be annoying to see one every time you want the other.

metadave:
ha, `:catalog` reminds me of my first days of programming: https://en.wikipedia.org/wiki/Apple_DOS

Trying to think as a student here, I think these are decent options:
- `catalog`
- `names`

There is also `:imports`, but perhaps it's too close to `import`.

byorgey:
Yeah, I thought of `:imports`, but I might expect that to list the names of the currently imported modules or something.  I think either `:names` and `:catalog` are fine, just pick whichever you like best!

PR #215, opened 2020-06-22
add :help repl command

metadave:
closes https://github.com/disco-lang/disco/issues/79


byorgey:
Happy to discuss this further via IRC or whatever if you want help fleshing out a design.

metadave:
I like the `Command` idea, I'll play around with the code for a bit and see what that looks like. Thanks @byorgey!

PR #217, opened 2020-06-23
Props & quantifiers

shaylew:
(Previous discussion at https://github.com/disco-lang/disco/pull/200.)


shaylew:
> Wow, it's amazing how much code we get to delete when switching to use `simple-enumeration`.

There could possibly be even more -- the `countType`, `isEmptyTy`, and `isSearchable` machinery is almost but not quite redundant with the value enumerations. Combining them would (practically) require putting all the enumeration code into Maybe so it could signal "nope, can't enumerate that", and (philosophically) introduce a weird dependency between the typechecker and the representation of terms in the evaluator, but those things might be worth it.

(There's also a tiny bit of code right now in `enumFunction` that works around a difference of opinion between Disco -- which is happy to enumerate `N -> Void` or `N -> Unit` -- and simple-enumeration, which complains about the infinite domain. I made an [issue](https://github.com/byorgey/enumeration/issues/1) there.)

> Ah, it's because nested quantifiers desugar to alternating quantifiers and cases:

Ah, oops, yeah. One quantifier with several bindings is desugaring differently from nested (matching) quantifiers, and the current attempt only works when there aren't intervening cases. 

I guess there are a couple levels of sophistication we could go for here:

1. The current behavior, where search works for one *syntactic* quantifier in the source program (and `exists x. exists y. p` is different from `exists x y. p`).
2. Lifting the abstraction out of the (single-branch) case, so syntactically-adjacent quantifiers are treated the same whether they're merged or broken up. We could either do this in the compiler (looking through cases) or the desugarer (merging the abstractions before we put the cases in). 
3. Doing something clever in the evaluator (keeping a quantifier stack?) to make the question of which quantifiers get searched together semantic rather than syntactic. For instance, no amount of desugarer magic is going to get this working, but it'd be slightly awkward if inlining `rangeIncludes` makes it work.
```
rangeIncludes : (B -> N) -> N -> Prop
rangeIncludes f y = exists (x : B). f x == y

!!! exists (f : B -> N). rangeIncludes f 10
```

byorgey:
> There could possibly be even more -- the countType, isEmptyTy, and isSearchable machinery is almost but not quite redundant with the value enumerations. Combining them would (practically) require putting all the enumeration code into Maybe so it could signal "nope, can't enumerate that", and (philosophically) introduce a weird dependency between the typechecker and the representation of terms in the evaluator, but those things might be worth it.

Interesting, I'd have to think about this more.  I don't like the sound of "weird dependency between the typechecker and the representation of terms in the evaluator" but I'm not sure I completely understand all the issues.  Let's definitely not worry about it for now and maybe we can clean things up even more down the road if we find a reason to.

> (There's also a tiny bit of code right now in enumFunction that works around a difference of opinion between Disco -- which is happy to enumerate N -> Void or N -> Unit -- and simple-enumeration, which complains about the infinite domain. I made an issue there.)

Oh, I wondered about that bit of code!  When I saw it I thought, "wait, doesn't `functionOf` already take care of that?"  Let's definitely fix that in `simple-enumeration`.

> The current behavior, where search works for one syntactic quantifier in the source program (and exists x. exists y. p is different from exists x y. p).

Ugh, I don't like this option at all.  I'd really like to be able to tell students that `exists x y z. ...` is an abbreviation for `exists x. exists y. exists z. ...`, and for them to be able to write nested quantifiers without weird gotchas about fair vs. unfair search.

> Lifting the abstraction out of the (single-branch) case, so syntactically-adjacent quantifiers are treated the same whether they're merged or broken up. We could either do this in the compiler (looking through cases) or the desugarer (merging the abstractions before we put the cases in).

Yeah, now that I think about it, I think the desugarer is the right place.  We already should change the desugarer so it avoids inserting redundant cases.  We might as well put in something to collect repeated quantifiers.  It is a bit odd though, it feels sort of like "sugaring" rather than "desugaring".

Option 3 is really interesting, and the example is compelling (if not inherently realistic).  I'm not sure it's worth doing right now (unless it turns out to be really easy?), but we should keep it in mind.

byorgey:
The Travis build was failing due to a Haddock parse error I created earlier.  Hopefully should be fixed now.

shaylew:
> It is a bit odd though, it feels sort of like "sugaring" rather than "desugaring".
I agree it's a little odd. If we were doing it just to lambdas it would feel more like a core-to-core optimization pass -- increasing the arity of a lambda to avoid immediately returning a closure, which incidentally makes it less strict. But doing it to quantifiers is less like an optimization with a minor semantic side effect and more like a semantic transformation with a minor performance benefit.

I thought a little more about option 3 and it seems like doing it in generality would require fair backtracking search. That sounds like a pain... except that `LogicT` exists right off the shelf! So while this would definitely be a complication, it does have the upside of giving a very solid semantics to values of type `Prop`: a prop would be (ignoring the example/counterexample) a `LogicT (Disco IErr) ()`, operations on props would combine those values, and `holds` would `runLogicT` them to observe whether they succeed or fail.

(That said, despite my excitement about this, it's probably better to just do (2) for now.)

byorgey:
> (That said, despite my excitement about this, it's probably better to just do (2) for now.)

Yeah, the `LogicT` thing sounds really cool, but I think I agree with your assessment.

byorgey:
`simple-enumeration-0.2.1` is now on Hackage with the enhancements to `functionOf`.

src/Disco/Desugar.hs
```diff
@@ -166,6 +167,13 @@ dtbin resTy p dt1 dt2
 mkDTPair :: DTerm -> DTerm -> DTerm
 mkDTPair dt1 dt2 = DTPair (getType dt1 :*: getType dt2) dt1 dt2
 
+------------------------------------------------------------
+-- Type helpers
+------------------------------------------------------------
+
+mkFunTy :: [Type] -> Type -> Type
```
byorgey, line 174: I'm pretty sure this function already exists somewhere in the Disco codebase...

byorgey:
How hard would it be to add another quantifier, `find`, such that `find (x1 : T1) ... (xn : Tn). prop` has type `Unit + T1 * ... * Tn`? *i.e.* like `exists` except it gives you the witness it finds instead of being a `Prop`.  For example
```
Disco> holds (exists (x:N) (y:N) (z:N). 0 < x < y < z and x^2 + y^2 == z^2)
true
Disco> find (x:N) (y:N) (z:N). 0 < x < y < z and x^2 + y^2 == z^2
right (3, 4, 5)
```

src/Disco/Desugar.hs
shaylew: In fact the place it exists (local to typechecking abstractions) is passing the answer right along, and in the place I though I needed to do this it had already been done. Maybe it belongs in `Types` but for now I just took it out and un-abstracted it back into `desugarDefn`.

shaylew:
I don't think that would be too hard. It's a new abstraction with its own typing rule, but on the evaluation end we're actually already generating the witness and then throwing it away.

byorgey:
@shaylew , let me know if there's any way I can help on this, or let me know if there are specific parts you're currently working on so I can pick a different aspect to work on.  For example, I could start working on getting rid of the special `Property` type and replace it with terms of type `Prop`.

shaylew:
I made a first pass at unifying propositions and properties. This pass is just the evaluator and not the frontend, so there's still a separate path for parsing and typechecking properties to be simplified away.

I think the major fiddly bit left is dealing with quantified argument names. Currently they get lost and replaced with deeply unhelpful (`arg0`, `arg11`) generated names at some point in the pipeline. And in fact it's not completely clear we want to report the argument names, rather than the pattern-bound variable names (though the one subsumes the other when applicable). If someone tries to test `forall (x: N, y: N). x < y or x > y` we'd maybe want to report a counterexamples with values for `x` and `y`, not values for the (unnamed) pair.

Maybe the thing to do here is to capture the environment at the point of failure, then somehow prune it to the variables of interest? Possibly sticking some more information on `VPSearch` values would make this easier.

The other (less fiddly) things to do are to recover "these things weren't equal" reporting with a Prop-valued equality operator, and to parse and typecheck properties as plain old expressions (of type Prop).

There are also definitely some opportunities for cleanup (e.g. finding a monoid for the expanded `TestResult` that's lazy enough for unbounded search), and some loose ends that may or may not pan out (e.g. the Prop negation stuff, which falls out for not-quite-free from the choice to represent `exists` and `forall` with one search value but which may or may not merit being exposed to the user).

shaylew:
@byorgey Do you have opinions about how to unify the syntax of properties and propositions? Right now properties don't require parenthesis around parameters but do require commas between them; but forall/exists in expressions work the other way around (like lambdas).

```
!!! forall a : Q, b : Q. divide a b * b == a

p :: Prop
p = forall (a : Q) (b: Q). divide a b * b == a
```

The former style is definitely closer to math notation, the latter is consistent with lambdas. There's no issue with supporting one or the other or both, it's just down to what syntax(es) you want to support on which abstractions.

byorgey:
Ah, good question, I had forgotten about this.  I definitely prefer the version with commas and no parentheses, for the reason you mention --- it is more in line with the usual mathematical syntax.

I think while we're at it we should just change the syntax of lambdas to match, i.e. `\x : N, y : Z. blah blah`.  I like not having to put parentheses around variables.  It feels weird that if you start out with `\x.x` but then you decide you want to put a type annotation on the `x`, you have to add parentheses.  Sure, it doesn't match Haskell or Agda syntax, but it will be more internally consistent and, I think, less confusing for students.

byorgey:
This is all looking good, by the way.  The issue with names for reporting counterexamples is tricky, I'd be happy to chat about that sometime if you want.

shaylew:
So `\x : N, y : Z. blah blah` would be a function of two arguments, and with parens `\(x : N, y : Z). blah blah` would take a tuple? And standalone definitions would still not use commas to keep them looking like their callsites?

```
foo : N -> Z -> Blah
foo x y = blah blah
```

byorgey:
Yes, I think so.  One thing to note here is that as a stylistic choice, Disco's standard library and educational materials are going to emphasize *uncurried* functions like
```
foo : N * Z -> Blah
foo(x,y) = blah blah
```
(because it is standard mathematical practice).  So I am not too worried about the inconsistency of requiring commas in multi-argument lambdas but not in multi-argument function declarations, because I think neither one of them is going to be very common.

shaylew:
Counterexamples are now working! I added a `:test <property>` command to the repl to make it easier to try out devious examples. I think the expressiveness of the Prop language has slightly outstripped the sophistication of the results reporting, but the messages are sensible if not perfect (and still work just as well on the kinds of properties you used to be able to write before).

Barring the inevitable finishing touches, I think this is more or less ready to come out of the oven.

src/Disco/Interactive/Eval.hs
```diff
@@ -227,46 +234,72 @@ runAllTests aprops
...
-  let qTys = M.fromList . fst . unsafeUnbind $ prop
-  prettyCounterexample qTys env
-prettyTestFailure prop (TestEqualityFailure ty v1 v2 env) = do
+  iputStr     "  - Test is false: " >> iputStrLn dp
+prettyFailureReason prop (TestEqual ty v1 v2) = do
   iputStr     "  - Test result mismatch for: "
   dp <- renderDoc $ prettyProperty (eraseProperty prop)
   iputStrLn dp
   iputStr     "    - Expected: " >> prettyValue ty v2
   iputStr     "    - But got:  " >> prettyValue ty v1
```
byorgey, line 271: I know this code was here from before, but while we are at it, maybe we should change this to say something like ` - LHS: ...\n - RHS: ...` The "expected/got" distinction is kind of arbitrary and not necessarily intuitive.

lib/prop.disco
```diff
@@ -0,0 +1,13 @@
+using Primitives
+
+||| Decide whether a proposition holds, *i.e.* convert it to a boolean.
+||| Note, however, that for some propositions this may run forever.
+||| For example, `holds (forall (x:N). x >= 0)` will run forever
+||| trying to check all natural numbers to see if they are all
+||| nonnegative.
+|||
+||| XXX what to do if you want to do randomized testing instead.
```
byorgey, line 9: We should fill in the `XXX` part.

src/Disco/AST/Core.hs
```diff
@@ -174,6 +179,13 @@ data Op = OAdd      -- ^ Addition (@+@)
         | OIsPrime        -- ^ Primality test
         | OFactor         -- ^ Factorization
 
+        -- Propositions
+        | OForall [Type]  -- ^ Universal quantification (function -> prop).
```
byorgey, line 183: What does `function -> prop` mean?

src/Disco/AST/Core.hs
```diff
@@ -174,6 +179,13 @@ data Op = OAdd      -- ^ Addition (@+@)
...
+        | OExists [Type]  -- ^ Existential quantification (function -> prop).
+        | OHolds          -- ^ Convert Prop -> Bool via exhaustive search.
+        | ONotProp        -- ^ Flip success and failure for a prop.
```
byorgey, line 186: Are we still using `ONotProp`?

src/Disco/AST/Core.hs
```diff
@@ -174,6 +179,13 @@ data Op = OAdd      -- ^ Addition (@+@)
...
+        | OShouldEq Type  -- ^ Equality assertion, @=!=@
```
byorgey, line 187: At this point it shouldn't be hard to add other types of binary assertions, right?  *e.g.* `!<`, `!<=`, etc.  We don't necessarily have to add them before merging, we could make a ticket and add them as a feature later.  Just asking what would be involved in adding them while I'm thinking of it.

src/Disco/AST/Desugared.hs
```diff
@@ -109,10 +115,16 @@ type instance X_TTup DS           = Void -- No tuples, only pairs
 type instance X_TParens DS        = Void -- No explicit parens
 
 -- Extra constructors
-type instance X_Term DS =
-  Either
-    (Type, DTerm, DTerm)                 -- DTPair
-    Type                                 -- DTCons
+type instance X_Term DS = X_DTerm
+
+data X_DTerm
```
byorgey, line 120: Oh, using an explicit data type instead of a nested `Either` is a really good idea.  Obvious in retrospect. =)

src/Disco/AST/Desugared.hs
```diff
@@ -154,11 +166,14 @@ pattern DTTyOp :: Type -> TyOp -> Type -> DTerm
 pattern DTTyOp ty1 tyop ty2 = TTyOp_ ty1 tyop ty2
 
 pattern DTNil :: Type -> DTerm
-pattern DTNil ty = XTerm_ (Right ty)
+pattern DTNil ty = XTerm_ (DTNil_ ty)
+
+pattern DTTest :: [(String, Type, Name DTerm)] -> DTerm -> DTerm
```
byorgey, line 171: Should maybe add a comment here saying what this AST node is used for.  (I *think* I know, but I am not 100% sure!)

src/Disco/AST/Generic.hs
```diff
@@ -323,13 +327,17 @@ data Term_ e where
...
   TString_ :: X_TString e -> [Char] -> Term_ e
 
-  -- | An anonymous function, /e.g./ @\x (y:N). 2x + y@.  There can be
-  --   multiple arguments, and each argument may be annotated with a
-  --   type.
-  TAbs_   :: X_TAbs e -> Bind (X_TAbsBind e) (Term_ e) -> Term_ e
+  -- | A binding abstraction, of the form @Q vars. expr@ where @Q@ is
+  --   a quantifier and @vars@ is a list of bound variables and
+  --   optional type annotations.  In particular, this could be a
+  --   lambda abstraction, /i.e./ an anonymous function (/e.g./ @\x
```
byorgey, line 333: We should update the syntax of the binder lists in this comment, e.g. `\x, y:N. 2x + y` instead of `\x (y:N). 2x + y`.

src/Disco/AST/Generic.hs
```diff
@@ -632,15 +640,34 @@ deriving instance ForallPattern Show         e => Show       (Pattern_ e)
...
+-- Quantifiers and binders
+------------------------------------------------------------
+
+-- | A type family specifying what the binder in an abstraction can be.
+--   Should have at least variables in it, but how many variables and
+--   what other information is carried along may vary.
+type family X_Binder e
+
+-- | A binder represents the stuff between the quantifier and the body
+--   of a lambda, ∀, or ∃ abstraction, as in (x y : N) (y : F).
```
byorgey, line 653: Should update this binder list syntax too.

src/Disco/AST/Generic.hs
```diff
@@ -632,15 +640,34 @@ deriving instance ForallPattern Show         e => Show       (Pattern_ e)
...
+
 ------------------------------------------------------------
 -- Property
 ------------------------------------------------------------
 
--- | A property is a list of forall-bound variables with their types,
---   followed by a (boolean) term.
+-- | A property is just a term (of type Prop).
 --
 --   This should probably move somewhere else?
```
byorgey, line 669: I think we can now delete this comment about this moving somewhere else.

In fact I wonder whether we even need to define `Property_` at all now.  Though I don't think it really hurts.

src/Disco/AST/Surface.hs
```diff
@@ -192,6 +194,8 @@ partitionDecls []                   = ([], [], [])
 ------------------------------------------------------------
 type Term = Term_ UD
 
+type instance X_Binder          UD = [Pattern]
```
byorgey, line 197: Might be good to add a comment here explaining what this definition allows. *e.g.* it is a list meaning that multiple things can be bound, and also noting that patterns can contain type annotations.

src/Disco/Extensions.hs
```diff
@@ -24,6 +24,7 @@ type ExtSet = Set Ext
 data Ext
   = Primitives   -- ^ Allow primitives, i.e. @$prim@
   | Randomness   -- ^ Allow randomness.  This is not implemented yet.
+  | QualifiedTypes  -- ^ Allow qualified types in surface syntax.
```
byorgey, line 27: We should remove this for now, it should definitely wait for a later PR.

src/Disco/AST/Core.hs
shaylew: We're neither exposing it nor desugaring in terms of it right now; I can take it out. I thought it might be nice to have (e.g. to explain that `not (forall (...). ...)` is the same as `exists (...). not (...)`) but it may not be worth it. It would cause a little more reporting complication to have it completely working, because it'd mean you can e.g. succeed because a boolean was false or fail because two things were equal.

src/Disco/Eval.hs
```diff
@@ -197,6 +207,78 @@ instance Show ValDelay where
...
+    --   then a number of additional cases were checked at random.
+  deriving Show
+
+-- | The answer (success or failure) we're searching for, and
+--   the result (success or failure) we return when we find it.
+--   The motive @(False, False)@ corresponds to a "forall" quantifier
+--   (look for a counterexample, fail if you find it) and the motive
+--   @(True, True)@ corresponds to "exists". The other values
+--   arise from negations.
+newtype SearchMotive = SearchMotive (Bool, Bool)
```
byorgey, line 228: Are we actually making use of both booleans, or is this left over from the stuff with a prim to invert a prop?  I tried tracing its use and it took me to `invertMotive`, which took me to `primNotProp`, which took me to `ONotProp`, which does not seem to be used at all.

If we are in fact not making use of this extra flexibility, should we get rid of it, or should we leave it in just in case we want it in the future?

src/Disco/Interactive/Eval.hs
```diff
@@ -227,46 +234,72 @@ runAllTests aprops
...
+prettyTestResult prop (TestResult _ r _)   = do
+    dp <- renderDoc $ prettyProperty (eraseProperty prop)
+    iputStr       "  - Test passed: " >> iputStrLn dp
+    prettySuccessReason r
+
+prettySuccessReason :: TestReason -> Disco IErr ()
+prettySuccessReason (TestFound (TestResult _ _ vs)) = do
+  prettyTestEnv "    Found example:" vs
+prettySuccessReason (TestNotFound Exhaustive) = do
+  iputStrLn     "    No counterexamples exist."
```
byorgey, line 255: I tried doing `:test forall x:B. not (not x) == x` and expected to get this message, but I got a different one instead:
```
Disco> :test forall x:B. not (not x) == x
  - Test passed: ∀x. ¬(¬x) == x
    Checked 100 possibilities without finding a counterexample.
```
Any idea why?

src/Disco/Interactive/Eval.hs
```diff
@@ -307,3 +340,13 @@ handleTypeCheck t = do
...
     Right (_,sig) -> renderDoc $ prettyTerm t <+> text ":" <+> prettyPolyTy sig
 
+evalTest :: Term -> Disco IErr ()
+evalTest t = do
+  ctx   <- use topCtx
+  tymap <- use topTyDefns
+  case evalTCM (extends ctx $ withTyDefns tymap $ checkProperty t) of
+    Left e   -> iprint e    -- XXX pretty-print
+    Right at -> do
+      r <- runTest 100 at   -- XXX make configurable
```
byorgey, line 350: This call to `runTest` also needs to happen within the scope of `extends ctx . withTyDefns tymap`:
```
Disco> isSquare = \n. exists x:N. x^2 == n
Disco> :type isSquare
isSquare : ℕ → Prop
Disco> isSquare 3
<Prop>
Disco> :test isSquare 3
Impossible! Unbound variable while interpreting: isSquare
CallStack (from HasCallStack):
  error, called at src/Disco/Interpret/Core.hs:231:16 in disco-0.1.0.0-IYCmSuepTQ6FMGPFQCyp7v:Disco.Interpret.Core
```

src/Disco/Interactive/Parser.hs
```diff
@@ -75,6 +76,7 @@ parseCommandArgs cmd = maybe badCmd snd $ find ((cmd `isPrefixOf`) . fst) parser
       , ("ann",     Ann       <$> term)
       , ("desugar", Desugar   <$> term)
       , ("compile", Compile   <$> term)
+      , ("test",    Test      <$> term)
```
byorgey, line 79: It will require a little bit of work to merge this with https://github.com/disco-lang/disco/pull/230 but it shouldn't be too hard.

src/Disco/Interpret/Core.hs
```diff
@@ -1134,6 +1154,87 @@ valueToString = fmap toString . rnfV
     toString (VCons 1 [VNum _ c, cs]) = chr (fromIntegral $ numerator c) : toString cs
     toString _ = "Impossible: valueToString.toString, non-list"
 
+------------------------------------------------------------
+-- Propositions
+------------------------------------------------------------
+
+------------------------------------------------------------
+-- NOTE: Reporting Counterexamples
+--
```
byorgey, line 1163: Was there going to be a note here?

src/Disco/Interpret/Core.hs
```diff
@@ -1134,6 +1154,87 @@ valueToString = fmap toString . rnfV
...
+  where
+    toProp b = VProp (VPDone (TestResult b (TestEqual t x y) []))
+
+-- | Convert a prop to a boolean by dropping its evidence.
+primHolds :: Value -> Disco IErr Value
+primHolds v = resultToBool =<< testProperty Exhaustive v
+  where
+    resultToBool :: TestResult -> Disco IErr Value
+    resultToBool (TestResult _ (TestRuntimeError e) _) = throwError e
+    resultToBool (TestResult b _ _) = return $ VCons (if b then 1 else 0) []
```
byorgey, line 1206: `if b then 1 else 0` is just `fromEnum b`.

src/Disco/Pretty.hs
```diff
@@ -433,11 +432,12 @@ prettyWHNF out (TyUser nm args) v = do
     Just (TyDefBody _ body) -> prettyWHNF out (body args) v
     Nothing                 -> error "Impossible! TyDef name does not exist in TyMap"
 prettyWHNF out TyUnit          (VCons 0 []) = out "()"
+prettyWHNF out TyProp          _            = prettyPlaceholder out TyProp
```
byorgey, line 435: It should be possible to do better here than just calling `prettyPlaceholder`.  But we should probably split it off into a separate issue to be addressed later.

src/Disco/AST/Core.hs
shaylew: Shouldn't be hard at all. The search code just deals with `TestResult`s without regard to their success/failure explanations, so it's just a matter of extending `TestReason`, extending the result printing code to explain the new reasons, and writing little ops in the style of `opShouldEq`.

src/Disco/Property.hs
```diff
@@ -15,257 +14,57 @@
...
+invertPropResult :: TestResult -> TestResult
+invertPropResult res@(TestResult b r env)
+  | TestRuntimeError _ <- r = res
+  | otherwise               = TestResult (not b) r env
+
+-- | Select samples from an enumeration according to a search type.
+generateSamples :: SearchType -> E.IEnumeration a -> Disco e [a]
+generateSamples Exhaustive e           = return $ E.enumerate e
+generateSamples (Randomized n m) e
+  | E.Finite k <- E.card e, k <= n + m = return $ E.enumerate e
```
byorgey, line 61: In another comment I asked why when I tested something of the form `forall x:B. ...` I was getting a message that said something about trying 100 possibilities.  Looking at this code, it seems like what is going on is that under the hood it *actually* just tried the two boolean values, but the `SearchType` value remained the same, and that's what the pretty-printer used to decide what message to display.  Maybe `generateSamples` should also return a potentially updated `SearchType` value??

src/Disco/AST/Core.hs
shaylew: It's the notional type of the op: `OForall [t1, t2, ... tn]` takes a closure of type `t1 -> t2 -> ... -> tn -> Prop` to a Prop. (I'll put that explanation in instead.)

It's possible going through Core via an op isn't even the right call, and there should just be a Core form for "quantified expression". It made more sense as an op back before I'd bitten the bullet and added new normal-form Values for props.

lib/prop.disco
shaylew: Is the idea to add a note that says "use `:test` at the repl for randomized testing" or to somehow expose randomized testing at the disco expression level?

src/Disco/Typecheck.hs
```diff
@@ -768,22 +774,41 @@ typecheck Infer             (TRat r)     = return $ ATRat r
...
+-- Lambdas and quantifiers are similar enough that we can share a
+-- bunch of the code, but their typing rules are a bit different.  In
+-- particular a lambda
+--
+--   \(x1:ty1) (x2:ty2) ... . body
+--
+-- is going to have a type like ty1 -> ty2 -> ... -> resTy, whereas a
+-- quantifier like
+--
+--   ∃(x1:ty1) (x2:ty2) ... . body
```
byorgey, line 788: We should update the binder syntax in this comment too.

test/parse-quantifiers/input
```diff
@@ -0,0 +1,35 @@
...
+:parse λ (x:N), y. x > y
+:parse λ (x:N), (y:F). x > y
+:parse λ x, (y:F). x > y
+:parse \ x. x > 3
+:parse \ (x:N). x > 3
+:parse \ (x:N), y. x > y
+:parse \ (x:N), (y:F). x > y
+:parse \ x, (y:F). x > y
+(\x, y. x + y) 3 5
+(\(x : N, y : N). x + y) 3 5
```
byorgey, line 12: Should we also add a succeeding version of this case, i.e. `(\(x:N,y:N).x+y) (3,5)`?

test/prop-basic/input
```diff
@@ -0,0 +1 @@
+:load test/prop-basic/prop-basic.disco
```
byorgey, line 1: Once we fix the problem with the evaluation context for `:test`, we should add some uses of these properties here with `:test`.

src/Disco/Eval.hs
shaylew: We're only using two values right now (exists is "succeed if you find a success" and forall is "fail if you find a failure"). The search code is slightly simplified by being able to abstract over "what am I looking for" and "what do I do when I find it", but without the NotProp stuff (which gives rise to the other two values) it'd make just as much sense to do a case and determine those locally.

Up to you if you think you're going to want negation, I think?

lib/prop.disco
byorgey: Just the former.  So someone reading the code will know about `holds` vs `:test`.

src/Disco/AST/Core.hs
byorgey: Oh, I see what you mean.  Yeah, a comment would be helpful.  As for whether this should exist as an op, I'm not sure.

src/Disco/Interactive/Eval.hs
shaylew: Ohh, this is actually a bit awkward. We're not actually testing 100 random Booleans -- see `generateSamples` in `Property.hs` -- but we're also not getting back a different SearchType to indicate that it's really exhaustive. This can be fixed.

src/Disco/AST/Core.hs
byorgey: OK, great.  I added https://github.com/disco-lang/disco/issues/242 to track this.

src/Disco/Eval.hs
byorgey: Just to clarify, when you say "if you're going to want negation", you mean a `Prop`-level negation, right?  I guess it would be nice to have if most of the machinery is already there. And actually we kind of do already have it, except not:
```
Disco> :type not (exists x:N. x == x)
not (∀(x : ℕ). x == x) : Prop
Disco> :test not (exists x:N. x == x)
user error (Pattern match failure in do expression at src/Disco/Interpret/Core.hs:372:3-12)
```

src/Disco/Pretty.hs
byorgey: Created https://github.com/disco-lang/disco/issues/243 to track this.

byorgey:
This is so cool!  I think it's looking really good.  After fixing the various minor things I mentioned in comments above, we should also (probably in this order):
1. Rebase on top of master and fix merge conflicts.
2. Fix compiler warnings.  There are a bunch at the moment, and I like to keep things `-Wall` clean as much as possible (see the `build` script for the GHC options we use).  There are a couple warnings about the pattern-match checker exceeding a certain number of iterations which are impossible to turn off, but other than that we should try to get rid of all warnings.  Happy to help with that if you want, let me know. 
3. Merge in the restyler PR (this can wait until the very end of course).

src/Disco/Interactive/Eval.hs
byorgey: Yeah, makes sense, you'll see I later left another comment on `generateSamples` making a similar suggestion.  I didn't realize the problem with nested samples though.

src/Disco/Interactive/Eval.hs
byorgey, line 350: The `100` should also ideally be configurable (this problem existed long before this branch!) so I also created https://github.com/disco-lang/disco/issues/244 to track that.

src/Disco/Interpret/Core.hs
shaylew: There is now!

test/prop-basic/input
shaylew: Done. It turns out randomized testing (this approach at least) is not very good at surjectivity, though -- the inner exists has to chance upon the right value, not just once but 100 times.

byorgey: Interesting.  So when I try `:test surjective(\x.x)` I get
```
  - No example was found: surjective (λx. x)
    Checked 100 possibilities.
```
So if I understand you correctly, you're saying that what we really want is for this to succeed and tell us it tried 100 examples for the outer forall and didn't find any counterexamples, hence it is likely to be true (clearly `holds` cannot be expected to terminate).  But it is exceedingly unlikely that the inner exists will get lucky 100 times and happen to pick the same value that was picked for the forall.

I don't really know how to do better in a principled way.  It might be a nice project for someone (e.g. a student in the future) to implement a sort of best-effort algebraic solver which could look at `exists x. x == 3` and just figure out what `x` should be rather than trying randomly (but could still fall back to random search in cases where the algebraic solver fails).

src/Disco/Eval.hs
shaylew: (Yeah, I meant prop-level negation.)

Oh huh, I didn't think that would typecheck! But in fact `not`, `and`, and `or` all check at `Prop -> Prop` as well as the expected `B -> B` and `B -> Prop`, unlike identical user-defined boolean functions!

In the case of `not` it'd be easy to not desugar it away and make one connective that works for either booleans or props, though I don't know if that's clearer or less clear than using a separate name for the prop version. But this works for `not` only because we can push it inside searches in the evaluator (by twiddling the motive). There's no representative in our type of ValProp normal forms that could represent the `and` or `or` of two searches.

byorgey: > But in fact not, and, and or all check at Prop -> Prop as well as the expected B -> B and B -> Prop, unlike identical user-defined boolean functions!

Yes, that's because of this code here: https://github.com/disco-lang/disco/pull/217/files#diff-6f4e13d554f1399d016fcb9dc348a635R477  I made those changes to the type checker intentionally when I started adding `Prop` but I hadn't really thought through all the implications yet.

Long-term, I think it would be ideal for `Prop` to support `and` and `or` (and possibly `not`), so one can start talking about proofs and evidence etc., maybe even connect `Disco` to some sort of external theorem provers and/or proof assistants, etc.  

I was in the middle of writing this when I saw your edit.  I was going to say that it seems clear at this point we should push it off to the future, put the type checking for `and`, `or`, `not` back the way they were, and create an issue to track this feature for the future.  So now I guess it's less clear, but maybe still a good idea.  What do you think?

shaylew: I think it's still a good idea to put it off. That trick gets the right answers but it's going to take some work to also get good error messages -- you really don't want to see "I tried 2 things and couldn't find an example" when you ask `:test false or false`.

byorgey: Right, makes sense.

test/prop-basic/input
shaylew: I don't know that it's reasonable to expect this to succeed, though it's too bad it doesn't. I could imagine a policy of trying more examples on the inner quantifier (the outer one magnifies the chance of error, so you want to drive the per-example error rate down lower), but it's probably not worth it.

(And ofc when qualified types happen the `surjective` in the standard library can just have documentation to the effect of "randomized testing won't be very effective when the domain is infinite".)

byorgey: Yeah, agreed.

byorgey:
This is looking good.  What's left to do before merging?

PR #222, opened 2020-06-26
help impl - compile cycle example

metadave:
```
Module imports form a cycle:
         module ‘Disco.Interactive.Parser’ (src/Disco/Interactive/Parser.hs)
        imports ‘Disco.Interactive.Commands’ (src/Disco/Interactive/Commands.hs)
  which imports ‘Disco.Interactive.Parser’ (src/Disco/Interactive/Parser.hs)
```

### DIFF
--- a/disco.cabal
+++ b/disco.cabal
@@ -244,6 +244,7 @@ library
                        Disco.Interactive.Eval
                        Disco.Interactive.Parser
                        Disco.Interactive.CmdLine
+                       Disco.Interactive.Commands
 
   other-modules:       Paths_disco
   autogen-modules:     Paths_disco
--- a/disco.cabal
+++ b/disco.cabal
@@ -245,6 +245,7 @@ library
                        Disco.Interactive.Parser
                        Disco.Interactive.CmdLine
                        Disco.Interactive.Commands
+                       Disco.Interactive.REPLCommand
 
   other-modules:       Paths_disco
   autogen-modules:     Paths_disco
--- a/src/Disco/Interactive/Commands.hs
+++ b/src/Disco/Interactive/Commands.hs
@@ -1,0 +1,70 @@
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Disco.Interactive.Comands
+-- Copyright   :  disco team and contributors
+-- Maintainer  :  byorgey@gmail.com
+--
+-- SPDX-License-Identifier: BSD-3-Clause
+--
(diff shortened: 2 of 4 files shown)


byorgey:
Sorry it's taken me longer than I hoped to get around to looking at this.  After drawing a bunch of dependency graphs on paper, I think I came up with something that will work.

- First, instead of having `parseCommandArgs` refer to `discoCommands` explicitly, just make it into a function that expects a list of type `[ReplCommand]` as an argument. `commandParser` will need an argument as well.  This is actually beneficial for more than just breaking the cycle; it makes it easy to programmatically provide different subsets of commands instead of having the list of commands baked in.
- Make a module `Disco.Interactive.Types` which contains the definitions of the `ReplCommandType`, `ReplCommand`, and `REPLExpr` types.
- Make a `Disco.Interactive.Parser` module which contains the things it currently contains, except that the definition of `REPLExpr` will be moved out, as well as the change to the type of `parseCommandArgs`.  It only has to import `Disco.Interactive.Types`.
- Make a module `Disco.Interactive.Commands` (which imports both `D.I.Types` and `D.I.Parser`) which defines the `discoCommands` list.
- Finally, `Disco.Interactive.Eval` can import both `Parser` and `Commands`, and can provide `discoCommands` (or any desired subset of it) as an argument to `commandParser`.

Let me know how this sounds. It's quite possible I've missed something.

metadave:
thanks @byorgey! I'll close this and try your suggestions out in the next few days.

metadave:
@byorgey the changes above seem to work ok, but getting the "action" of each ReplCommand to work introduces an additional circular dep:

```haskell
-- module Disco.Interactive.Types
data ReplCommand = ReplCommand
  { name  :: String 
  , shortHelp :: String
  , longHelp :: String
  , cmdType :: ReplCommandType
  , cmdAction :: REPLExpr -> Disco IErr ()  -- the action that occurs when you type (for example) :help
  , cmdParser :: Parser REPLExpr
  }

-- module Disco.Interactive.Commands
ReplCommand {
    name = "type",
    shortHelp = "Typecheck a term",
    longHelp = "Typecheck a term",
    cmdType = User,
    cmdAction = handleTypeCheck t >>= iputStrLn, -- reference to Disco.Interactive.Eval
    cmdParser = TypeCheck <$> parseTypeTarget
}

-- module Disco.Interactive.Eval
import Disco.Interactive.Commands <---- circular dep


```

PR #226, opened 2020-07-02
refactor REPL commands to implement :help

metadave:
You mentioned in #215 that `The tricky part of this is that different commands need different numbers and types of arguments, so actually including the actions to be run for each command would require some type trickery.` What I have here might not be exactly what you're looking for, but it's my naive first attempt :-) I'm happy to rework the implementation, it was a fun learning experience.

A `REPLCommand` action is of type `REPLExpr -> Disco IErr ()`, so you'll see wrapper functions that accept this type, unwrap it and pass it to the more specific function (`handleAnn` in this case, which is of type `Term -> Disco IErr String`). The sad part is the need to match all the other cases and `return ()`. Maybe this isn't ideal, but it was easy enough to implement. 

```
handleAnnWrapper :: REPLExpr -> Disco IErr ()
handleAnnWrapper (Ann t) = handleAnn t >>= iputStrLn
handleAnnWrapper _       = return ()

handleAnn :: Term -> Disco IErr String
handleAnn t = do
```

TODO:

- [ ] add a test for :help
- [ ] rebase on several commits from July 1st


byorgey:
@metadave yes, this is exactly what I had in mind when I said this would be the tricky part.  It's not very satisfying to have all the `handleXXXWrapper` functions which match on one constructor of `REPLExpr` and `return ()` on everything else.  I just opened a PR illustrating how to do it in a more satisfying (though more complicated!) way with type-indexed GADTs.  Right now I only got the `:ann` command working as an illustration and everything else is still commented out, so it would still need a lot of cleanup and converting the other commands.

What do you think?  Happy to explain more if you're unsure how it works.

Issue #185, opened 2020-06-02
Be a bit more lax when parsing :type command


byorgey:
It would be nice to be able to say
```
> :type max
~max~ : N x N -> N
```
At the moment, though, you get a parse error since `max` is a reserved word and can't be used as an identifier.  Should be able to add a special case just for this.

byorgey:
Actually can't do this in a reasonable way until we represent all operators as their own standalone syntax node + function application, see #136 .

Issue #136, opened 2018-08-07
User-defined infix operators


byorgey:
It would be really cool to allow users to define infix operators (especially in terms of moving stuff into the standard library).  While we're at it we might allow postfix operators too.  There are a number of tricky things to decide here but in the end it should be doable.  Here's my current thinking.

First, we need some syntax to denote when a name is supposed to be an infix (postfix) operator, and so it can be used independently (*i.e.* so it can be partially applied, or so one can ask for its type using `:type`, etc.).  The best idea I have currently is to do what Agda does and use underscores for this purpose.  That is, disallow underscores in names, and use something like `_foo_` to denote the name of an infix operator `foo`. The idea is that the underscores stand in place of the arguments.

Next, there needs to be a way to declare the associativity and precedence level of an infix operator.  Haskell allows such "fixity" declarations to occur anywhere in a file, but this makes parsing tricky.  To make this work Haskell requires infix operators and prefix functions to be *syntactically* distinguished (operators use only operator symbols, prefix functions use alphanumerics).  That way, a module can be parsed using a "default" fixity for any unrecognized operators, and then once all the fixity declarations have been collected, a second pass goes back through and fixes up any expressions to reflect the true fixities.

However, we have already gone down the path of allowing alphanumeric names as infix operators (*e.g.* `mod`, `choose`) and I think that fits best with the design philosophy of the language.  So allowing fixity declarations throughout a module is not really an option (unless we disallowed infix operators from being used at all in the same module where they are defined, but this seems terrible).  I think a good compromise is to require all fixity declarations for a module to come *first* (after any imports but before any other declarations).  Then after parsing them we can dynamically construct an expression parser and use it to parse the rest of the module.

What should a fixity declaration look like?  Haskell has `infix`, `infixl`, or `infixr` to specify no, left, or right associativity, followed by a number from 0-9 to indicate precedence level.  However, we already have a lot more than 9 precedence levels, and they aren't numbered in any particular way (the `Syntax.Operators` module assigns them numbers arbitrarily based on their order in the table, just to be able to compare levels easily).  I propose a way to specify associativity, just like Haskell, and then a way to specify another operator for comparison and whether the new operator's precedence is lower, higher, or equal to the other operator.  Then it will get inserted in the operator table appropriately.

For example, suppose we want to define an operator `*!` which performs matrix multiplication, and we want it to be left associative and have the same precedence level as normal multiplication.  We might write something like
```
infix _*!_ same precedence as _*_, left associative
```
[...]

PR #228, opened 2020-07-04
get rid of TUn and TBin AST nodes completely

byorgey:
Unary and binary operations were getting desugared into applications
of operator primitives, but still initially parsed as separate special
TUn and TBin nodes.  This patch gets rid of the special TUn and TBin
nodes and now parses unary and binary operators directly as a
TApp (TPrim ...) ...  TUn and TBin are now pattern synonyms, so a lot
of code that used them does not have to change.

This is a preliminary step towards #185 and #136.


byorgey:
@shaylew can you see if this makes sense to you?

src/Disco/Parser.hs
```diff
@@ -877,12 +877,6 @@ parseExpr = (fixJuxtMul . fixChains) <$> (makeExprParser parseAtom table <?> "ex
     fixPrec (TBin bop1 t1 (TBin bop2 t2 t3))
       | bPrec bop2 < bPrec bop1 = TBin bop2 (fixPrec $ TBin bop1 t1 t2) t3
 
-    fixPrec (TApp (TBin bop t1 t2) t3)
-      | bPrec bop < funPrec = TBin bop t1 (fixPrec $ TApp t2 t3)
-
-    fixPrec (TApp t1 (TBin bop t2 t3))
-      | bPrec bop < funPrec = TBin bop (fixPrec $ TApp t1 t2) t3
-
```
shaylew, line 885: Is the idea here that nothing will bind tighter than functions (in which case these are unreachable)? This makes sense, but it may also mean there's no reason for a configurable `funPrec` in the first place -- after this the parser won't use it, though the pretty printer still does.

byorgey:
Oh, very good catch re: the dead code in the pretty-printer.  In fact, you caught a bug: you're right that the two code paths for unary operators (depending on whether it was a `TApp TPrim` or a `TUn`) were different, and in fact it turns out that factorials (our only postfix unary operator at the moment) were being pretty-printed incorrectly, e.g. `:pretty 5!` would show `!5`.  But the code for `TApp TPrim` already handled this correctly. So I added a test and deleted the dead code.

src/Disco/Parser.hs
byorgey: I *think* this code was actually unreachable before, for exactly the reason you mention, but I had never noticed.  However, after making `TUn` and `TBin` into pattern synonyms, the second one became reachable!  In particular the second case matches when a lower-precedence unary operator is applied to a higher-precedence binary operator, and there's exactly one case where that can sensibly happen, when an exponentiation is negated.  In fact the reason I noticed is that precisely one expression in one single test started failing: -3^2 was evaluating as 9 instead of -9.   It took me a long time to track it down!

You may be right about not needing a configurable `funPrec`, I'll have to think about it more.

byorgey: Actually I think `funPrec` is fine.  It is definitely needed in the pretty-printer. And it's not so much that it's configurable, as that it's automatically assigned to be one level higher than the highest operator.  If we ever add new operators with a precedence level in between existing ones, the `funPrec` will be automatically updated appropriately.

byorgey:
Thanks!

PR #229, opened 2020-07-04
show how to use GADTs and type indices to dispatch commands

byorgey:
This way we don't need the `toTag stuff`, and we don't need any
`handleXXXWrappe`r functions that match on just one thing and `return ()`
for everything else.

The key idea is that both commands and and REPLExprs are indexed by an
enumeration, so each REPLExpr and the command that can process it
share a label.  The 'dispatch' function uses the magic 'gcast'
function from Data.Typeable to find the first command which has a
matching type index, and uses its 'action' to process
the (guaranteed-matching) REPLExpr.  Each 'handleXXX' function
declares that it only handles REPLExprs with a particular type index,
which means it only has to match on the one constructor which actually
has that type index.

### DIFF
--- a/src/Disco/Interactive/Commands.hs
+++ b/src/Disco/Interactive/Commands.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE GADTs     #-}
 -----------------------------------------------------------------------------
 -- |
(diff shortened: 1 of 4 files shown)


metadave:
this is really interesting, and it looks like I have some reading to do :-)

I'll merge into my wip branch and refactor all the rest. Thanks @byorgey!

PR #230, opened 2020-07-06
refactor repl commands, add :help command

metadave:
GADT/typed indices implementation from @byorgey via https://github.com/disco-lang/disco/pull/229

- [x] decide on :help output
- [x] add test for :help

see also:
- https://github.com/disco-lang/disco/pull/226


src/Disco/Interactive/Commands.hs
```diff
@@ -0,0 +1,560 @@
...
+typeCheckCmd :: REPLCommand 'CTypeCheck
+typeCheckCmd =
+    REPLCommand {
+        name = "type",
+        shortHelp = "Typecheck a term",
+        longHelp = "",
+        category = Dev,
+        cmdtype = ShellCmd,
+        action = handleTypeCheck,
+        parser = TypeCheck <$> parseTypeTarget
```
byorgey, line 280: Can we also move the definition of `parseTypeTarget` to right here?  Unless I am mistaken this is the only place it is used.

src/Disco/Interactive/Commands.hs
```diff
@@ -0,0 +1,560 @@
...
+        garbageCollect
+
+handleHelp :: REPLExpr 'CHelp -> Disco IErr ()
+handleHelp Help = do
+  iputStrLn "Commands available from the prompt:\n"
+  mapM_ (\(SomeCmd c) -> iputStrLn $ ":" ++ name c ++ "\t\t" ++ shortHelp c) $  sortedList discoCommands
+  iputStrLn ""
+  where
+    sortedList cmds = sortBy (\(SomeCmd x) (SomeCmd y) -> compare (name x) (name y)) $ filteredCommands cmds
+    -- remove "builtins" (let, eval, etc), don't show dev-only commands by default
```
byorgey, line 364: It would be nice if we could find a good way to show built-in commands too, though I think that can probably be put off to the future.

metadave: @byorgey I added build-in commands to the help output.

src/Disco/Interactive/Commands.hs
```diff
@@ -91,342 +91,321 @@ annCmd :: REPLCommand 'CAnn
...
+handleImport (Import modName) = catchAndPrintErrors () $ do
+  mi <- loadDiscoModule FromCwdOrStdlib modName
+  addModule mi
+
+
 letCmd :: REPLCommand 'CLet
 letCmd =
     REPLCommand {
       name = "let",
+      helpcmd = "let <expression>",
```
byorgey, line 252: Should be `let <variable> = <expression>`.

src/Disco/Interactive/Commands.hs
```diff
@@ -450,6 +442,18 @@ handleTypeCheck (TypeCheck t) = do
         Right (_,sig) -> renderDoc $ prettyTerm t <+> text ":" <+> prettyPolyTy sig
   iputStrLn s
 
+usingCmd :: REPLCommand 'CUsing
+usingCmd =
+    REPLCommand {
+        name = "using",
+        helpcmd = ":using <extension>",
```
byorgey, line 449: Shouldn't have a colon in front of `:using`.

metadave: fixed, is `using` a dev-only command?

src/Disco/Interactive/Commands.hs
```diff
@@ -91,342 +91,321 @@ annCmd :: REPLCommand 'CAnn
...
+      io . putStrLn $ p
+      case M.lookup x docs of
+        Just (DocString ss : _) -> io . putStrLn $ "\n" ++ unlines ss
+        _                       -> return ()
+
 evalCmd :: REPLCommand 'CEval
 evalCmd =
     REPLCommand {
       name = "eval",
+      helpcmd = "<eval>\t\t",
```
byorgey, line 181: Let's say `<term>` instead of `<eval>`.

src/Disco/Interactive/Commands.hs
```diff
@@ -91,342 +91,321 @@ annCmd :: REPLCommand 'CAnn
 annCmd =
     REPLCommand {
       name = "ann",
+      helpcmd = ":ann",
```
byorgey, line 94: I like the idea of having a `helpcmd` field.  However, let's not put explicit tab characters in them --- both because we should avoid using tabs in the output (different terminals etc. may display them differently) and also because it's fragile.  Instead let's make the code to display the help more general, so it figures out the maximum length of all the `helpcmd` fields it's being asked to display, and pads things out with spaces approriately.

src/Disco/Interactive/Commands.hs
byorgey: No, `using` is for end users to turn on language extensions.

src/Disco/Interactive/Commands.hs
byorgey: Great!

src/Disco/Interactive/Commands.hs
metadave: agreed on tabs, I'll push another change to this PR soonish.

metadave:
@byorgey squashed commits, removed tabs, padding `:help` columns with strings

test/repl-help/input
```diff
@@ -0,0 +1,2 @@
+:help
+1+1
```
metadave, line 2: Restyled.io wants to strip the whitespace after the `:help` command, so I added a simple `1+1` to produce additional output.

byorgey:
Looks good to me!

PR #239, opened 2020-07-14
Allow naked operators in :type command

### LINKED ISSUE #185: Be a bit more lax when parsing :type command

byorgey:
For example, one can now write `:type max` (which used to yield a
syntax error); it behaves identically to `:type ~max~`.

Closes #185.


byorgey:
Thanks for taking a look!  I'll merge it but in general you're welcome to merge in this situation too.

Issue #80, opened 2018-04-16
Add logs for other bases


byorgey:
For example, right now there is no way to compute the number of decimal digits in a number, since we only have log base 2.  (Note that `lg n / lg 10` does not give the correct result since `lg` rounds down.)

Not sure what the syntax should be.

byorgey:
We don't need special syntax; we can just make a primitive `$log : N * N -> N` and expose it via a library function, probably in `num.disco`.  In fact while we're at it, we can get rid of built-in syntax for `lg` and just define it in the library too, as `$log(2,n)`.

ericpinter:
At the moment it seems like there is a little bit more going on with how `lg` works. An expression like `lg 0` returns a built in error `LgOfZero`. Is it worth keeping this behavior or is a simple `NonExhaustive` enough?

byorgey:
Yes, `NonExhaustive` is more for user-defined functions that encounter a missing case at runtime.  For built-in things like `log` we should have more specific errors when they encounter undefined behavior.

PR #251, opened 2020-09-19
Integer Base Logarithms

ericpinter:
Solves #80 by adding logarithms of integer base as a library function. Also removes everything I could find from the language where base-2 lg was built in.


byorgey:
BTW, for future reference, if you use the magic words "closes" or "fixes" (as in `closes #80` or `fixes #80`) then GitHub automatically links them so that merging this PR will close #80.  Unfortunately "solves" is not a magic word. =)

byorgey:
Also for future reference, since I've given you push access to this repo, everything tends to work more smoothly if instead of opening a PR from your own fork (`ericpinter:master`), you make a new branch in this repo (e.g. something like `disco-lang:log-base`) and open a PR from there.

byorgey:
If you look at the "failing checks" below you'll see that one of the test cases is failing.  Looks like one of the test cases uses `lg` which is no longer built-in.  It might be as simple as adding an `import` statement to the test.

lib/num.disco
```diff
@@ -9,6 +9,24 @@ using Primitives
...
 isPrime = $isPrime
 
+!!! log (1,1) == 0
+!!! log (3,1) == 0
+!!! log (2, 1024) == 10
+!!! log (2, 1023) == 9
+!!! log (7, 11398895185373143) == 19	
+log : N * N -> N
+log (1,1) = 0
+log (0,_) = $crash "Log with this base is undefined"
```
byorgey, line 19: The error message could be more specific: "Log base 0 is undefined" or something like that.  And similarly for the one below.

lib/num.disco
```diff
@@ -9,6 +9,24 @@ using Primitives
 isPrime : N -> Bool
 isPrime = $isPrime
 
+!!! log (1,1) == 0
```
byorgey, line 12: Should also include a documentation comment (starting with `|||`).

lib/num.disco
```diff
@@ -9,6 +9,24 @@ using Primitives
...
+!!! log (2, 1024) == 10
+!!! log (2, 1023) == 9
+!!! log (7, 11398895185373143) == 19	
+log : N * N -> N
+log (1,1) = 0
+log (0,_) = $crash "Log with this base is undefined"
+log (1,_) = $crash "Log with this base is undefined"
+log (_,0) = $crash "Log of Zero"
+log (b, x) = {?
+				0					if b > x,
```
byorgey, line 23: Looks like you used some tab characters here which makes it look misaligned depending on how tab characters are displayed.  Let's replace the tab characters with spaces (you can also look into configuring your editor of choice so that it automatically replaces tabs with spaces in `.disco` files).

byorgey (changes requested):

byorgey:
Can you just fix the restyled thing?  It only made this change re-adding a newline at the end of a file which one of your commits removed: https://github.com/disco-lang/disco/pull/252/commits/1d5e9c88cd5c8b8cc9539f453dff5184b4e3285c    If you add that newline back then https://github.com/disco-lang/disco/pull/252 should automatically close itself, then I'll merge this one.

byorgey (approved):

byorgey:
Oh, I almost forgot: the other thing we should do is make sure the tutorial is up to date, probably just by removing any mention of `lg`.

PR #256, opened 2021-03-22
Graphs

byorgey:
Creating a pull request as a place to discuss and eventually to merge this code into the main branch.


src/Disco/Parser.hs
```diff
@@ -936,10 +936,11 @@ parseAtomicType = label "type" $
...
 parseCon :: Parser Con
 parseCon =
-      CList <$  reserved "List"
-  <|> CBag  <$  reserved "Bag"
-  <|> CSet  <$  reserved "Set"
-  <|> CUser <$> parseTyDef
+      CList  <$  reserved "List"
+  <|> CBag   <$  reserved "Bag"
+  <|> CSet   <$  reserved "Set"
+  <|> CGraph <$  reserved "Graph"
```
byorgey, line 942: It looks like we forgot to add `Map` to the parser?

src/Disco/Pretty.hs
```diff
@@ -447,6 +451,11 @@ prettyWHNF out (TySet t) (VBag xs) =
   out "{" >> prettySequence out t (map fst xs) ", " >> out "}"
 prettyWHNF out (TyBag t) (VBag xs) = prettyBag out t xs
 
+prettyWHNF out (TyGraph a) (VGraph g adj) = prettyWHNF out (TyMap a (TySet a)) =<< rnfV adj
+prettyWHNF out (TyMap k v) (VMap m) = prettyWHNF out (TySet (k :*: v)) =<< mapToSet k v (VMap m)
+prettyWHNF out (TyGraph a) (VConst (OGEmpty _)) = out $ "[]"
+prettyWHNF out (TyMap k v) (VConst OEmpty)  = out $ show (M.empty :: M.Map SimpleValue SimpleValue)
```
byorgey, line 457: Since the argument to `prettyWHNF` is guaranteed to be in WHNF already, I don't think we can possibly get either of these `VConst` cases.

src/Disco/Pretty.hs
```diff
@@ -447,6 +451,11 @@ prettyWHNF out (TySet t) (VBag xs) =
   out "{" >> prettySequence out t (map fst xs) ", " >> out "}"
 prettyWHNF out (TyBag t) (VBag xs) = prettyBag out t xs
 
+prettyWHNF out (TyGraph a) (VGraph g adj) = prettyWHNF out (TyMap a (TySet a)) =<< rnfV adj
+prettyWHNF out (TyMap k v) (VMap m) = prettyWHNF out (TySet (k :*: v)) =<< mapToSet k v (VMap m)
```
byorgey, line 455: As we've discussed, we should consider pretty-printing graphs and maps in such a way that the output is valid Disco code which could be used to reconstruct the value, e.g. by adding a call to some kind of conversion function on the front of the output.

src/Disco/Pretty.hs
```diff
@@ -540,6 +549,12 @@ prettyDecimal r = printedDecimal
          (pre, rep) = splitAt len expan
          first102   = take 102 expan
 
+--prettyGraph :: (String -> Disco IErr ()) -> Type -> Value -> Disco IErr ()
+--prettyGraph out ty val = out "Graph <" >> go val
+--    where 
+--      go (VGraph g) =
```
byorgey, line 555: We can delete this commented-out code.

src/Disco/Types.hs
```diff
@@ -463,7 +479,7 @@ pattern TyString = TyList TyC
 
 {-# COMPLETE
       TyVar, TySkolem, TyVoid, TyUnit, TyBool, TyProp, TyN, TyZ, TyF, TyQ, TyC,
-      (:->:), (:*:), (:+:), TyList, TyBag, TySet, TyUser #-}
+      (:->:), (:*:), (:+:), TyList, TyBag, TySet, TyGraph, TyUser #-}
```
byorgey, line 483: Need to add `TyMap` to this `COMPLETE` pragma as well.

src/Disco/Types.hs
```diff
@@ -550,6 +566,13 @@ countType (TyBag ty)
   | isEmptyTy ty        = Just 1
   | otherwise           = Nothing
 countType (TySet ty)    = (2^) <$> countType ty
+--as far as I can tell the number of graphs constructable with a label type that counts to t
+--is the sum from n=0 to t of (t choose n)*(4^(n*(n-1)/2))
+--(t choose n) is the number of ways we can choose n distinct labels
+--(4^(n*(n-1)/2)) represents deciding for each pair of nodes whether there is an edge pointing a->b, b->a, both, or neither
+--this function diverges ridiculously quickly though, so until further revision this is omitted
+--countType (TyGraph ty)  = Nothing
+countType (TyMap tyKey tyValue) = (\a b -> (b+1) ^ a) <$> countType tyKey <*> countType tyValue
```
byorgey, line 576: I think this needs a couple additional special cases. If either `k` or `v` is empty, then `Map k v` has size 1 (even if the other type is infinite), i.e. the empty map is the only value.

src/Disco/Types.hs
```diff
@@ -550,6 +566,13 @@ countType (TyBag ty)
   | isEmptyTy ty        = Just 1
   | otherwise           = Nothing
 countType (TySet ty)    = (2^) <$> countType ty
+--as far as I can tell the number of graphs constructable with a label type that counts to t
+--is the sum from n=0 to t of (t choose n)*(4^(n*(n-1)/2))
+--(t choose n) is the number of ways we can choose n distinct labels
+--(4^(n*(n-1)/2)) represents deciding for each pair of nodes whether there is an edge pointing a->b, b->a, both, or neither
```
byorgey, line 572: I think this doesn't properly account for self-loops, which are allowed.  I think we can simplify things and account for self-loops at the same time with  sum_{n=0}^t (t choose n) 2^(n^2)

i.e. for each ordered pair of vertices (u,v) (including the possibility that u = v), we choose whether or not there is a directed edge u -> v.

It's in OEIS, but the fact that the entry doesn't mention graphs at all makes me suspicious I did something wrong. https://oeis.org/A135748

byorgey: Oh, I know why: because it's a bit strange to count the number of graphs whose labels are a *subset* of a given set of labels.  But we have a good reason to.  Maybe we should submit a comment to that OEIS entry. =)

byorgey: We might as well implement this.  We already have binomial coefficients implemented somewhere (maybe we just call out to a library, I forget), so it should be one line of code.

byorgey (commented):
We should also add a bunch of tests to the test suite.

PR #271, opened 2021-06-18
Update version bound in cabal

TristanCacqueray:
This change enables building the project with latest cabal
registry.


TristanCacqueray:
This is no longer necessary.

PR #272, opened 2021-06-18
show user defined types with the :names command

### LINKED ISSUE #259: :names command should list user-defined types too
Right now `:names` only lists value-level bindings; we should include type definitions too.

TristanCacqueray:
Fixes #259

### DIFF
--- a/src/Disco/Interactive/Commands.hs
+++ b/src/Disco/Interactive/Commands.hs
@@ -325,8 +325,13 @@ namesCmd =
 handleNames :: REPLExpr 'CNames -> Disco IErr ()
 handleNames Names = do
   ctx  <- use topCtx
+  tyDef <- use topTyDefns
+  mapM_ showTyDef $ M.toList tyDef
   mapM_ showFn $ M.toList ctx
   where
+      showTyDef (name, body) = do
+        p <- renderDoc (prettyTyDef name body)
+        io . putStrLn $ p
       showFn (x, ty) = do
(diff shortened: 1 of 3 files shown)


src/Disco/Interactive/Commands.hs
```diff
@@ -325,8 +325,13 @@ namesCmd =
 handleNames :: REPLExpr 'CNames -> Disco IErr ()
 handleNames Names = do
   ctx  <- use topCtx
+  tyDef <- use topTyDefns
+  mapM_ showTyDef $ M.toList tyDef
   mapM_ showFn $ M.toList ctx
   where
+      showTyDef (name, body) = do
+        p <- renderDoc . hsep $ [prettyTyDef name body]
```
byorgey, line 333: Since there is only one thing in the list, the `hsep` and the list is actually unnecessary, we can just say `renderDoc (prettyTyDef ...)`

byorgey:
Looks great so far!  We should also add a test case.  Probably we can just add it to `test/repl-names`.  Just add a type definition or two (with & without parameters) to `logic.disco` there and update the `expected` file accordingly.`

src/Disco/Interactive/Commands.hs
TristanCacqueray: Oh right.

byorgey:
Thanks!  The CI failure was my fault and should be fixed now.

PR #273, opened 2021-06-18
document the ~ notation

### LINKED ISSUE #236: Write about ~+~ operator forms in the tutorial
If e.g. `+` is a binary operator, one can write `~+~` to refer to it as itself (e.g. to pass it to a higher-order function, to ask for its type, etc.).  For example:
```
Disco> :type ~+~
~+~ : ℕ × ℕ → ℕ
Disco> import list
Loading list.disco...
Disco> foldr(~+~,0,[1 .. 10])
55
```
In general, one puts a `~` anywhere an argument is expected.  So e.g. `~+~ and ~*~ but also ~! (factorial) and -~ (negation)`. Currently I do not think this is mentioned in the tutorial.

TristanCacqueray:
Fixes #236

### DIFF
--- a/docs/tutorial/functions.rst
+++ b/docs/tutorial/functions.rst
@@ -167,3 +167,27 @@ not a multiplicative term.
    :math:`x(y+3)` as function application, although they might also
    rightly complain that :math:`x` is a strange choice for the name of
    a function.
+
+Operator functions
+==================
+
+Operators can be manipulated as function using the ``~`` notation.
+The tilde goes wherever the argument to the operator would go.
+This can be used to pass them to a higher-order function.
+
+::
+    Disco> :type ~+~
+    ~+~ : ℕ × ℕ → ℕ
+
+    Disco> import list
+    Loading list.disco...
+    Disco> foldr(~+~,0,[1 .. 10])
+    55
+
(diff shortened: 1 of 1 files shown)


docs/tutorial/functions.rst
```diff
@@ -167,3 +167,21 @@ not a multiplicative term.
...
+    Disco> :type ~+~
+    ~+~ : ℕ × ℕ → ℕ
+    Disco> import list
+    Loading list.disco...
+    Disco> foldr(~+~,0,[1 .. 10])
+    55
+
+.. note::
+
+    For unary operator, only one ``~`` is needed, e.g. ``~!`` for negation.
```
byorgey, line 187: `~!` is actually factorial, not negation.

TristanCacqueray: Oops :) So what is the logic for the position of the tild, should not negation be `~-` ?

byorgey: Negation is `-~`.  The tilde goes wherever the argument to the operator would go.  So postfix operators go after the tilde, and prefix operators go before the tilde.

byorgey:
Thanks!

PR #274, opened 2021-06-18
enumerate for infinite types

### LINKED ISSUE #91: enumerate for infinite types?
Right now, `count` and `enumerate` both require a finite type.  For `count` this makes sense.  For `enumerate`, however, there's no reason in principle that it couldn't return a (lazy, infinite) list of all the values of an infinite type.

cdsmith:
Enumeration already works for infinite types, but the primitive
previously had a check to fail when the type was infinite.  This
just removes that check.

Fixes #91

### DIFF
--- a/src/Disco/Eval.hs
+++ b/src/Disco/Eval.hs
@@ -500,6 +500,13 @@ initDiscoState = DiscoState
 type Disco e = StateT (DiscoState e) (ReaderT Env (ExceptT e (LFreshMT IO)))
 
 ------------------------------------------------------------
+-- Some needed instances.
+
+-- This should eventually move into unbound-generics.
+instance MonadFail m => MonadFail (LFreshMT m) where
+  fail = LFreshMT . Fail.fail
+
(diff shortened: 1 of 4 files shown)


src/Disco/Eval.hs
```diff
@@ -499,13 +499,6 @@ initDiscoState = DiscoState
 --   * Can do I/O
 type Disco e = StateT (DiscoState e) (ReaderT Env (ExceptT e (LFreshMT IO)))
 
-------------------------------------------------------------
--- Some needed instances.
-
--- This should eventually move into unbound-generics.
-instance MonadFail m => MonadFail (LFreshMT m) where
-  fail = LFreshMT . Fail.fail
-
```
byorgey, line 508: Can you remove this deletion from the PR?  This orphan instance is still needed to build with `stack`.

cdsmith: Done.

byorgey:
Thanks!

PR #301, opened 2021-07-25
Fix Variable not in scope

amir:
Currently `stack build` fails with Variable not in scope error. Adding `mapMaybe` to imports.


byorgey:
Thanks!  Not sure how that slipped through...

PR #355, opened 2022-11-08
Propositions Under AND and OR

cally-cmd:
Added functionality for testing propositions under binary operators AND and OR


src/Disco/Property.hs
```diff
@@ -116,7 +116,14 @@ prettyFailureReason
   => TestReason -> Sem r Doc
 prettyFailureReason TestBool = empty
 prettyFailureReason (TestEqual ty v1 v2) =
```
byorgey, line 118: We will need to add a `AProperty` argument to both `prettyFailureReason` and `prettySuccessReason` so that they can pattern-match on it in the `TestAnd` and `TestOr` cases and pass on sub-props to recursive calls appropriately.

src/Disco/Property.hs
```diff
@@ -12,6 +12,8 @@
 -----------------------------------------------------------------------------
 
 {-# LANGUAGE OverloadedStrings #-}
+{-# OPTIONS_GHC -Wno-incomplete-patterns #-}
+{-# OPTIONS_GHC -Wno-unused-matches #-}
```
byorgey, line 16: These are useful warnings, if they are being generated then we should fix the cause rather than disabling them.

src/Disco/Property.hs
```diff
@@ -28,13 +30,17 @@ module Disco.Property
 
 import           Prelude                     hiding ((<>))
 
+import           Debug.Trace
```
byorgey, line 33: Should delete this import.

src/Disco/Value.hs
```diff
@@ -360,21 +361,33 @@ resultIsCertain (TestNotFound (Randomized _ _)) = False
...
     c1 = testIsCertain tr1
     c2 = testIsCertain tr2
-    -- ok1 = testIsOk tr1
-    -- ok2 = testIsOk tr2
+    ok1 = testIsOk tr1
+    ok2 = testIsOk tr2
+resultIsCertain (TestImpl tr1 tr2)
+  | c1 && c2      = True
+  | not ok1 && c1 = True
+  | otherwise     = False
```
byorgey, line 386: What about adding `ok2 && c2`?  If you know `q` is true then you know `p -> q` is true regardless of `p`.

byorgey (commented):
This is looking pretty good!  We should definitely add a bunch of unit tests.  Let me know if you need help figuring out how to do that.

For some reason `:test` sometimes prints out just a single bullet with no content, but I'm not sure why:
```
Disco> :test forall n:N. (n < 10) /\ (n < 8)
  -
```

src/Disco/Property.hs
```diff
@@ -22,19 +24,23 @@ module Disco.Property
...
   )
        where
 
 import           Prelude                     hiding ((<>))
 
+import           Debug.Trace
+
 import qualified Data.Enumeration.Invertible as E
-import qualified Test.QuickCheck             as QC
+-- import qualified Test.QuickCheck             as QC
```
byorgey, line 36: We should clean up this commented-out import.

src/Disco/Property.hs
```diff
@@ -66,11 +77,15 @@ generateSamples (Randomized n m) e
   | E.Finite k <- E.card e, k <= n + m = return (E.enumerate e, Exhaustive)
   | otherwise                          = do
     let small = [0 .. n]
-    rs <- runGen . mapM sizedNat $ [n .. n + m]
+    -- we don't have to worry about getting too large a number since we handle finite lengths in the case above.
+    -- rs <- runGen . mapM sizedNat $ [n .. n + m]
+    -- in this moment, we want 50 random numbers such that the first one is in [n, n^2],
+    -- the second one is in [n^2, n^3] ... etc.
```
byorgey, line 83: We should update these comments, I'm not sure they are still accurate.

src/Disco/Property.hs
```diff
@@ -66,11 +77,15 @@ generateSamples (Randomized n m) e
...
+    -- rs <- runGen . mapM sizedNat $ [n .. n + m]
+    -- in this moment, we want 50 random numbers such that the first one is in [n, n^2],
+    -- the second one is in [n^2, n^3] ... etc. 
+    rs <- randomLarge [100, 1000, 10000, 100000, 1000000]
     let samples = map (E.select e) $ small ++ rs
     return (samples, Randomized n m)
-  where
-    sizedNat k = QC.resize (fromIntegral k) QC.arbitrarySizedNatural
+  -- where
+  --   sizedNat k = QC.resize (fromIntegral k) QC.arbitrarySizedNatural
```
byorgey, line 88: Clean up commented-out code.

src/Disco/Property.hs
```diff
@@ -82,55 +97,61 @@ prettyResultCertainty :: Members '[LFresh, Reader PA] r => TestReason -> AProper
...
-  prettyResultCertainty r prop "false"
-  $+$
-  prettyFailureReason r
+  => Bool -> AProperty -> TestReason -> Sem r Doc
+prettyTestReason b _ TestBool = empty
+prettyTestReason b prop (TestFound (TestResult bool tr env))
+  | b = prettyTestEnv "Found example:" env
+  | not b = prettyTestReason b prop tr $+$ prettyTestEnv "Found counterexample:" env
+prettyTestReason b _ (TestNotFound Exhaustive)
+  | b = "No counterexamples exist"
```
byorgey, line 108: Let's add `all possible values were checked` to this message, just like the one in the `not b` case.

src/Disco/Value.hs
```diff
@@ -312,10 +312,16 @@ data TestReason_ a
...
+  | TestLt Type a a
+    -- ^ The test was a less than test. Records the values being 
+    --   compared and also their type (which is needed for printing).
   | TestNotFound SearchType
     -- ^ The search didn't find any examples/counterexamples.
   | TestFound TestResult
     -- ^ The search found an example/counterexample.
+  | TestAnd TestResult TestResult
+  | TestOr  TestResult TestResult
+  | TestImpl TestResult TestResult
```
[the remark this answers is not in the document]
byorgey: Oh, hmm, never mind, `resultIsCertain` would break.

byorgey (commented):
This is coming along very well!  I have a few comments on some things requiring minor cleanup (and a couple comments from before that still need to be addressed).  The only big thing still missing is that we need to add some tests to the test suite (check out `test/README.md` and let me know if you have questions).

byorgey:
Closing in favor of rebased/cleaned up version #365 .

PR #380, opened 2024-05-20
Drop dependency on `polysemy-zoo`

### LINKED ISSUE #375: Drop dependency on `polysemy-zoo`
`polysemy-zoo` hasn't been updated in a while and only builds with GHC <= 9.4.  We should probably see about removing it as a dependency.

LeitMoth:
Copied over ConstraintAbsorber, MonadCatch, and Random from `polysemy-zoo`, so that we can remove it as a dependency. Included the `polysemy-zoo` license.
Removed Disco's `Random.hs` file, which only re-exported `polysemy-zoo`'s Random
Updated the needed imports.


LeitMoth:
I don't think putting the issue number in the name of the pull request properly referenced it.
This should resolve #375.

LeitMoth:
It also seems that the exact same thing was already done on the ghc-9.8 branch:
f72a7dbbd418e89593e674009ea17c1c79af6bfe
So this work may have been redundant.

byorgey:
> It also seems that the exact same thing was already done on the ghc-9.8 branch:

Hah, oops, I had totally forgotten about that.  That's why I was getting deja vu during our meeting this morning.  That's OK, we can merge this and I will update the `ghc-9.8` branch.

byorgey:
@LeitMoth Note you'll need to click on `Details` next to the failed `restyled` check below and follow the instructions for applying the restyling, then push the new commit.  The code copied into our repository needs to conform to our style guidelines.

PR #385, opened 2024-05-23
Fixed crash when 0^(-1)

### LINKED ISSUE #379: Zero to a negative exponent crashes Disco
**Describe the bug**
Trying to evaluate an expression such as `0^(-1)` causes Disco to crash.

**To Reproduce**
```
Disco> 0^(-1)
disco: Ratio has zero denominator
```

**Expected vs actual behavior**
Obviously this should be an error, since `0^(-1) = 1/(0^1) = 1/0`.  However, it should give a "division by zero" error just like actually dividing by zero does, not cause disco to crash entirely.
```
Disco> 1/0
Error: division by zero.
Disco> 0^(-1)
disco: Ratio has zero denominator
```

justingrubbs:
Fixes #379


byorgey (approved):

byorgey (changes requested):
Looks good!  Let's add some tests before merging this.

byorgey (approved):

PR #388, opened 2024-05-25
Display multiple example types for things with nontrivial (non-parametric) polymorphism

### LINKED ISSUE #169: Display multiple example types to the user in situations where we do not necessarily have a most general type
In order to always be able to infer a most general type, we would have to expose a lot more complex features in the surface-level type system (qualified polymorphism, subtyping constraints, ...).  The resulting types would be long and complex.  So we don't do that.  At the moment we can infer parametrically polymorphic types (e.g. `\x.x` is inferred to have type `a -> a`), but if there are any constraints at all (qualifiers, subtype constraints...) on a type variable then we monomorphize it before showing the result to the user.  This is *sound* but it leads to some unexpected behavior.  For example:
```
Disco> :type \x. x - 3
λx. x - 3 : ℤ → ℤ
Disco> :type (\x. x - 3) (1/2)
(λx. x - 3) (1 / 2) : ℚ
```
First, `\x. x - 3` is inferred to have type `ℤ → ℤ`, which is certainly a valid type for it.  It may be surprising, then, that the second example type checks at all: surely we are not allowed to apply a function expecting `ℤ` to a fraction?  And indeed, if we define our own function explicitly of type `ℤ → ℤ` then we can't:
```
Disco> :type let f : Z -> Z = \x. x - 3 in f (1/2)
Unsolvable NoUnify
```
What's actually going on here is that `\x. x - 3` is internally inferred to have a type like `(N <: a, QSub a) => a -> a`, that is, it has type `a -> a` for any `a` which supports subtraction and is a supertype of `N`; however, we can't show that to the user so we monomorphize it first, arbitrarily choosing `Z` for `a`, although `Q` also would have been a valid choice.  However, typechecking the expression `(\x. x - 3) (1/2)` generates an additional constraint `F <: a`, and in that case `Q` is the only valid choice for `a`.

So in the case of `:type \x. x - 3` I'm imagining showing the user something like this instead:
```
Disco> :type \x. x - 3
λx. x - 3 : (ℤ → ℤ) OR (ℚ → ℚ)
```
(or maybe putting each type on a separate line, or some other nice way of showing them).

In an ideal world, I would love if we could show the user a single type ONLY when we are sure it is the most general type or the only possible type.  I think this should be possible if we are conservative; I don't know if it is possible to decide in general (that is, there may be situations where we show multiple monomorphized types even though there is some single most general type we *could have* shown?).  To do a better job picking most general monomorphic types when possible we will definitely want to do something like #163 .

The other big caveat is that it's east to come up with expressions having a very large number of potential monomorphized types. For example, `\a1. \a2. \a3. ... \an. a1 + a2 + a3 + ... + an` has something like `4^n` different types, depending on which specific numeric types we pick for each of the arguments.  So in general we would probably need to cut it off at some point and just show a few options with a `...` at the end, or something like that.

byorgey:
Closes #169.  Previously, we did this:
```
Disco> :type \x. x - 2
λx. x - 2 : ℤ → ℤ
```
However, this was a lie since in fact `\x. x - 2` is more general than that; but showing only one possible monomorphic type was still better than showing some generic monstrosity like `forall a. (N <: a, subtractive a) => a -> a`. 

Now, with this PR, we do this:
```
Disco> :type \x. x - 2
This expression has multiple possible types.  Some examples:
λx. x - 2 : ℤ → ℤ
λx. x - 2 : ℚ → ℚ
```
See #169 for more context and examples.

Briefly, in this PR, we:
- Change the solver to return a *list* of solution substitutions instead of just a single one
- Add a "solution limit" counter to the solver so it can stop early once it has hit the limit for the number of solutions we want.
    - In some cases (e.g. when checking a provided type) we just want one.
    - In other cases (when inferring a type) we set an arbitrary bound of 16.
    - This is not really ideal --- in theory we would want to use some kind of proper backtracking logic monad like `LogicT` but that seems difficult/impossible to incorporate into `polysemy` (https://stackoverflow.com/questions/62627695/running-the-nondet-effect-once-in-polysemy; https://github.com/polysemy-research/polysemy/issues/246).
- "Thin" the resulting list of solutions by throwing away any solutions which are supertypes of another solution in the list.
- Add pretty-printing that distinguishes between a single solution and multiple solutions

It is still possible to make the solver blow up in case there are many possible container variables to instantiate.  *e.g.* expressions like `\x. \y. \z. \w. (set(x), set(y), set(z), set(w))` take a very long time to typecheck.  I have some ideas for how to improve this situation, but for now it's an uncommon corner case.


byorgey:
@justingrubbs , @LeitMoth I'd love for you to take a look at this when you get a chance (no rush).  I certainly don't expect you to read and understand all the code changes, though you're welcome to look at it in whatever detail you like.  But see if the overall idea makes sense, play with some examples, see if you can break it, that kind of thing.

Issue #55, opened 2017-07-26
Randomness


byorgey:
It would be nice to incorporate randomness into the language somehow.  There is a large design space to think about.

1. We could simply make the language semantics nondeterministic, so something of type (say) `N` might yield a different result every time the program is run.  This would be simple: we could just introduce some primitives for generating random values, with no real changes to the type system required at all.
2. On the other hand, it might be nice to distinguish deterministic from non-deterministic computations via types.  But it should still be possible to write things like `(3 + random)` --- students should not be forced to use `do`-notation or something similar to work in the "`Random` monad", for example.  Perhaps some of the recent work on effect types could be relevant here (e.g. https://arxiv.org/pdf/1406.2061.pdf).
    * We could use a `Random` type constructor which is a monad.  The potential problem is that we would want `t` to be a subtype of `Random t`, but that would mean that subtyping is no longer purely structural, which would probably make a bunch of things more complicated.  Another downside of this approach is that if we have something of type, say, `Random N`, there is no way to tell whether it is a computation which will generate a new random number every time it is evaluated, or a particular number which happened to be chosen via randomness.
    * We could also use an effect type on function arrows; the potential problem here is that this may only work if the language is call-by-value, since then function application is the only place where an effect may happen.  Though it may be that for randomness we don't actually care that much exactly *when* the random values are chosen --- this is different than, say, an exception or I/O effect.

byorgey:
After thinking about this a lot more, I think I have come to the conclusion that THE big problem is *how to control when sampling happens*.  Consider the example of wanting a list of random numbers.  We might imagine code something like
```
[ uniform (0,99) | x <- [1 .. 10] ]
```
which we suppose might give us a list of ten random numbers between 0 and 99.  But consider something like
```
let y = uniform (0,99) in [ y | x <- [1 .. 10] ]
```
This *should* be the same---substituting a let-binding shouldn't change the semantics!---but we could imagine it instead meaning a list of ten copies of the *same* (randomly chosen) number.  And this latter semantics might actually be something we want.

One could certainly imagine formalizing the exact evaluation semantics and specifying that sampling happens at evaluation time, so that the two expressions above *do* yield different results.  But this makes the language seem fairly imperative, and it's quite difficult to specify it in a way that's understandable in the presence of laziness.

It goes deeper than this; even with the above idea there are things we might want that we can't express.  In particular we can't build new distributions that yield different results every time we use them.  For example
```
dist = uniform (0,9) + uniform (0,9)
[...]

PR #390, opened 2024-06-03
Implements barebones randomness

justingrubbs:
Begins to address #55. 

```
Disco> random((1,10000),seed 10000)
(3200, <Gen>)
Disco> random((1,10000),seed 12)
(7611, <Gen>)
```

In this PR, we:
- Add primitive seed and random constructs that allow for the threading of a pseudorandom number generator through computations
- Modify example RSA file to generate random prime numbers to be used for encryption/decryption
- Add test file that shows computations are still deterministic with randomness


test/prim-random/prim-random.disco
```diff
@@ -0,0 +1,94 @@
+import num
+import list
+
+-- Implementation of RSA encryption algorithm.
+-- Reference: https://simple.wikipedia.org/wiki/RSA_(algorithm)
+
+-- To use, first call `getKeys` with two prime numbers, which returns
```
byorgey, line 7: Let's update this so it explains how to call `randKeys` instead of `getKeys`.

test/prim-random/prim-random.disco
```diff
@@ -0,0 +1,94 @@
...
+-- two pairs. The first pair is the public key, the second is the
+-- private key. These keys, along with the `encrypt` and `decrypt`
+-- functions can be used to encrypt and decrypt lists of natural
+-- numbers.
+
+encrypt : N * N -> List(N) -> List(N)
```
byorgey, line 13: It would be cool to update `encrypt` and `decrypt` to work on `List(Char)` instead of `List(N)`, but that should probably be separate from this PR.

byorgey (approved):
Looks great!

byorgey:
@justingrubbs You can update the comment in `rsa.disco` and then once you're happy with it, go ahead and click the "Squash and merge" button (you might have to select it from the dropdown if the button doesn't already say "Squash and merge").

test/prim-random/prim-random.disco
byorgey: Hmm, never mind, it looks like we never actually added primitive `Char <-> N` conversion functions, so we would have to add those first.

PR #394, opened 2024-07-06
Memoize

justingrubbs:
In this PR, we:

- Add automatic memoization of _most_ functions
- Include an example disco file, `example/fib.disco`, that demonstrates this automatic memoization as well as some of its limits


src/Disco/Compile.hs
```diff
@@ -182,11 +184,47 @@ compileDTerm term@(DTAbs q _ _) = do
...
+  abstract xs body = CAbs False (bind (map coerce xs) body)
+
+  abstractMemo :: [Name DTerm] -> Core -> Core
+  abstractMemo xs body = CAbs True (bind (map coerce xs) body)
 
   quantify :: Op -> Core -> Core
   quantify op = CApp (CConst op)
 
+  -- Intend on changing this, but functional for now
+  canMemo :: [Type] -> Bool
```
byorgey, line 196: `canMemo` looks like it is equivalent to `canMemo = all canMemoTy` for some appropriate `canMemoTy :: Type -> Bool`.

src/Disco/Compile.hs
```diff
@@ -182,11 +184,47 @@ compileDTerm term@(DTAbs q _ _) = do
...
+
+  checkAtom :: Atom -> Bool
+  checkAtom (AVar _) = False
+  checkAtom (ABase b) = checkBase b
+
+  checkBase :: BaseTy -> Bool
+  checkBase CtrList = False
+  checkBase CtrBag = False
+  checkBase CtrSet = False
+  checkBase Gen = False
```
byorgey, line 224: `P` (proposition) `BaseTy` should also not be memoizable.

disco.cabal
```diff
@@ -486,6 +486,7 @@ library
                        polysemy-plugin >= 0.4 && < 0.5,
                        reflection >= 2.1.7 && < 2.2,
                        random >= 1.2.1.1 && < 1.3,
+                       list-tries,
```
byorgey, line 489: I'm still not sure I understand what issues you were running into that were solved by using this package, and it's cool that you were able to find it and figure it out.  In theory it should not be necessary though, and I'd rather avoid depending on this package (it hasn't been updated since 2020 and already does not allow the most recent version of the `containers` library).

src/Disco/Compile.hs
```diff
@@ -182,11 +184,47 @@ compileDTerm term@(DTAbs q _ _) = do
...
 
+  -- Intend on changing this, but functional for now
+  canMemo :: [Type] -> Bool
+  canMemo [] = True
+  canMemo (x : xs) = case x of
+   TyAtom a -> checkAtom a && canMemo xs
+   TyCon CArr tys -> arrMemo tys && canMemo tys && canMemo xs
+   TyCon c tys -> checkCon c && canMemo tys && canMemo xs
+
+  arrMemo :: [Type] -> Bool
```
byorgey, line 203: `arrMemo` does not look necessary.  Essentially we are just making sure there are no arrow types.  It should be easy to write `canMemoTy :: Type -> Bool` to be `False` on arrow types, then we won't need a separate `arrMemo` function.

src/Disco/Compile.hs
```diff
@@ -181,12 +183,46 @@ compileDTerm term@(DTAbs q _ _) = do
...
 
   quantify :: Op -> Core -> Core
   quantify op = CApp (CConst op)
 
+  canMemo :: [Type] -> Bool 
+  canMemo = all canMemoTy
+
+  canMemoTy :: Type -> Bool 
+  canMemoTy (TyAtom a) = canMemoAtom a 
+  -- Anti-higher-order while permitting multiple arrows
```
byorgey, line 197: This doesn't seem right.  Multiple arrows should already be taken care of by the call to `unbindDeep`, so we end up with an uncurried `Lam` that takes a list of arguments; we then call `all canMemoTy` on that list of argument types.  So at this point we should not allow any arrows at all.  I think you can just get rid of this special case for `TyCon CArr` and let it be handled by the generic `TyCon` case.

src/Disco/Compile.hs
```diff
@@ -181,12 +183,46 @@ compileDTerm term@(DTAbs q _ _) = do
...
 
-  abstract :: [Name DTerm] -> Core -> Core
-  abstract xs body = CAbs (bind (map coerce xs) body)
+  abstract :: ShouldMemo -> [Name DTerm] -> Core -> Core
+  abstract m xs body = CAbs m (bind (map coerce xs) body)
 
   quantify :: Op -> Core -> Core
   quantify op = CApp (CConst op)
 
+  canMemo :: [Type] -> Bool
```
byorgey, line 192: You should add some comments to these functions explaining what they do and justifying some of the choices.

src/Disco/Interpret/CESK.hs
```diff
@@ -113,6 +113,8 @@ data Frame
     FUpdate Int
   | -- | Record the results of a test.
     FTest TestVars Env
+  | -- | Memoize the result of a function.
```
byorgey, line 116: Expand this comment to explain what the `Int` and `SimpleValue` represent.

src/Disco/Interpret/CESK.hs
```diff
@@ -194,15 +202,33 @@ step cesk = case cesk of
...
+   Just (n,mem) -> do
+      let sv = toSimpleValue $ foldr VPair VUnit (v : mem)
+      mv <- memoLookup n sv
+      case mv of
+         Nothing -> return $ In b (Ctx.insert (localName x) v e) (FMemo n sv : k)
+         Just v' -> return $ Out v' k
+
+  (Out v (FApp (VClo mi e (x : xs) b) : k)) -> case mi of
+   Just (n,mem) -> return $ Out (VClo (Just (n,v:mem)) (Ctx.insert (localName x) v e) xs b) k
+   Nothing -> return $ Out (VClo mi (Ctx.insert (localName x) v e) xs b) k
```
byorgey, line 221: Again, we want to factor out the parts that are the same and make the `case` as local as possible.  In this case it should be possible to even do without an explicit `case`, like so:
```suggestion
  (Out v (FApp (VClo mi e (x : xs) b) : k)) -> 
    return $ Out (VClo (second (v:) <$> mi) (Ctx.insert (localName x) v e) xs b) k
```

byorgey (commented):
Fantastic work!  Just a few more minor things to address.

byorgey:
Also, be sure to update the PR description to accurately describe the changes here.  The PR description will turn into the commit description once we merge this.

src/Disco/Compile.hs
justingrubbs: When I trace the list of argument types after the call to `unbindDeep`, there are still some `CArr` present. I initially had no special case for `TyCon CArr` but that resulted in nothing being memoized, including the basic `fib` function.

byorgey: Hmm, strange, let me think about this a bit more...

byorgey:
For the Restyled check, go to https://github.com/disco-lang/disco/pull/395 and click the "Squash + merge" button.

src/Disco/Compile.hs
byorgey: OK, it looks like we are inconsistent in how we use the `Type` field of `DTAbs`, which means `unbindDeep` is returning somewhat nonsensical results.  Let me keep thinking about the best way forward.

byorgey: I haven't taken the time to look at this more deeply yet.  Let's go ahead and merge this, and I'll create a new issue about cleaning up the code with respect to `DTAbs`/`unbindDeep` etc.

byorgey (approved):
Looks great!

PR #397, opened 2024-07-18
Updates to reference documentation

byorgey:
* New documentation about collection types
* New documentation about REPL commands
* Various minor fixes and cleanup

PR #405, opened 2024-10-26
Some updates + additions to the reference documentation

byorgey:
Enhanced or added reference documentation on:

- REPL commands
- containers (bags, Boom hierarchy)
- comprehensions
- `each` function
- a bunch of other stuff

As soon as we merge this it means the documentation on readthedocs will be updated, so *in theory* we shouldn't merge this until making a new release; but I'd rather just get things merged.  No one will really be looking at the docs right now.

### DIFF
--- a/docs/reference/.#string.rst
+++ b/docs/reference/.#string.rst
@@ -1,0 +1,1 @@
+brent@diophantus.10955:1721342881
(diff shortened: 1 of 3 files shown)


byorgey:
Looks like I got confused and this was already merged in #397 .

Issue #442, opened 2025-02-05
More precise type for `.-` operator


byorgey:
I received this question from a student:
```
-- btw shouldn't the negative numeric overloads return a positive type?
-- ~.-~ : ℕ × ℕ → ℕ
-- ~.-~ : ℤ × ℤ → ℕ
-- ~.-~ : 𝔽 × 𝔽 → 𝔽
-- ~.-~ : ℚ × ℚ → 𝔽
-- I don't see a reason why "4 .- -1" should be ℤ if it has a "lower bound of 0"
```
Indeed, this seems reasonable.  Right now we give `.-` a type like `a * a -> a`, the same as the other arithmetic operators:
```
 inferPrim (PrimBOp op) | op `elem` [Add, Mul, Sub, Div, SSub] = do
    a <- freshTy
    constraint $ CQual (bopQual op) a
    return $ a :*: a :->: a
```
but we could do something similar as for the type of absolute value.

PR #444, opened 2025-04-23
More precise type-inferencing for Saturating Subtraction

JCT03:
Saturating subtraction was separated from other arithmetic operators for more precise type inferencing. The operation can only return a zero or positive result, and the type should reflect that ([#442](https://github.com/disco-lang/disco/issues/442)). The implementation leverages the function created for absolute value types cAbs.


JCT03:
Yes, I will do that. I also see there are some existing tests for saturating subtraction that will I verify.

byorgey:
GitHub won't even let us merge the PR until all the tests pass, so technically there's no need to verify existing tests, but checking them manually doesn't hurt.

JCT03:
I added tests for the type of a saturating subtraction. 

Also, note that one downside of this implementation is the type of error produced for a command like `[1] .- [1]` changes. Previously, such a command would result in the error 

> Error: values of type List(a1) cannot be added and multiplied.
https://disco-lang.readthedocs.io/en/latest/reference/not-qual.html

. Now, the resulting error is 

> Error: the shape of two types does not match.
https://disco-lang.readthedocs.io/en/latest/reference/shape-mismatch.html

. The original error is informative, even if somewhat inaccurate for saturating subtraction. The new error, however, is not informative because it does not really get at the issue of the argument type being inappropriate for saturating subtraction.

byorgey:
I just pushed a new commit to main fixing the CI, but it looks like you'll need to rebase your PR on top of that new commit in order for the automated checks to work.  Let me know if you already know how to do that or if you'd like some help.

JCT03:
Ok, I believe I have successfully merged the main branch into my pull request, but please let me know if there is anything else I need to do.

byorgey (approved):
Looks good to me!

byorgey:
I'm really not sure what's wrong with the Restyled check, I'll have to look into it later.  Let's just ignore it for now.

